PR #51, opened 2026-09-19
chore(infra): scaffold the package, the gate script and CI

Ollienel777:
Closes #4

Opened by the build loop.

## What

- `pyproject.toml`: package `whiteout` (Python 3.11), deps `numpy`, `scipy`,
  `networkx`; `[dev]` extra `ruff`, `mypy`, `pytest`, `build`, plus
  `setuptools>=70.1` (the gate's `--no-isolation` build needs the backend
  present in the environment rather than fetched from PyPI). `ruff` and `mypy`
  are `==`-pinned: their output is the gate's pass/fail signal.
  `.claude/**` is excluded from ruff, mypy and pytest; `slow` is registered in
  `[tool.pytest.ini_options] markers`, and `--strict-markers` is on so an
  unregistered mark is an error rather than a silent no-op.
- `whiteout/` with `cli.py` exposing `run|replay|score|sweep|ablate|serve`.
  `run` and `score` are minimal but real, so the gate's smoke and determinism
  steps exercise the exact command lines from `SPEC.md` §6: `run` writes a
  placeholder JSONL determined by `--seed`, `--ticks` and `WHITEOUT_TRANSPORT`
  (the header record carries the transport as provenance), `score`
  reads a weights file and prints the four axes. The other four subcommands
  exit 1 with "not implemented yet". `WHITEOUT_SEED` is resolved inside
  `cmd_run`, not as an argparse default, so a non-integer value is a
  diagnosed `run: WHITEOUT_SEED=... is not an integer` and exit 1 rather than
  a traceback out of every subcommand; `score` likewise diagnoses a
  malformed or non-numeric weights file instead of raising.
- `scripts/gate.py`: the eight steps of `SPEC.md` §6 in order, failing fast,
  with each command copied from the spec's table, including
  `pytest -q -m "not slow"` and `python -m build --wheel --no-isolation`. It
  forces `WHITEOUT_TRANSPORT=kinematic`, opens no socket, prunes both
  `build/` and `dist/` before building, and prints one
  `GATE PASS:`/`GATE FAIL:` line per step.
- `.github/workflows/ci.yml`: one `gate` job on every pull request with no
  `paths:`/`paths-ignore:` filter, and on pushes to `main` — checkout,
  `setup-python@v5` (3.11), `pip install -e ".[dev]"`, `python scripts/gate.py`.
  `cancel-in-progress` is scoped to `pull_request`: on push the concurrency
  group is `refs/heads/main`, shared by every merge, so cancelling there would
[...]

Issue #25, opened 2026-09-19
feat: the episode loop, with a ≥100× real-time budget


Ollienel777:
**Milestone:** M1 · **Demo beat:** 5 · **Size:** M

## What
Wire terrain, vehicles, sensors and target into `whiteout/sim/episode.py`,
vectorised with NumPy, writing the episode log. Measure and report the
throughput of a 6-vehicle, 256×256, 400-tick episode against the ≥100×
simulated-wall-clock budget.

**The throughput assertion is marked `-m slow` and is excluded from the gate**
(`SPEC.md` §6). A wall-clock threshold measured on shared CI runners and from
parallel worktrees flakes, and a red gate on an unrelated PR is a PR that
cannot settle. It is asserted once here and re-measured by the tuner (#31).

## Why
The overnight parameter search is the entire strategic bet and it is impossible
without this. Measuring it here, as a `-m slow` test outside the gate that the
tuner re-measures, means we learn at hour 6, not hour 26.

## Acceptance
- [ ] `python -m whiteout.cli run --seed 7 --ticks 400` writes a valid log
- [ ] Two runs at the same seed produce byte-identical logs
- [ ] The throughput number is measured and reported in the PR
- [ ] The ≥100× assertion exists as a `-m slow` test, passes locally, and is
      not run by `scripts/gate.py`
- [ ] No module-level `np.random` anywhere in `whiteout/sim/`

## Notes
`SPEC.md` §5 "Conventions", §6, §10. Depends on #8, #10, #11. Run **#24 (S2)**
first.

Ollienel777:
loop: carry-forward from PR #52's review — route the CLI through the writer

Round 4 of PR #52 (the episode log) flagged this while verifying something
else, and it lands on this ticket rather than on that PR:

> `whiteout/cli.py:89` writes its episode log with `out.write_text(...)`,
> bypassing `write_episode_log`. It is the #51 scaffold placeholder
> (`"schema": "placeholder"`), not in this PR's diff, but the ticket that
> wires the CLI should route it through the writer or the promise has a hole
> outside `log.py`.

**Why it matters.** PR #52 spent four rounds establishing one property: every
path that replaces an episode log on disk must produce something the reader
accepts. Four separate defects of that class were found and closed — a
truncating open, an unvalidated payload, a reader-only monotonic check, and a
zero-record write. The writer now round-trips every record through the reader
before `os.replace`, and `os.replace` occurs exactly once in the tree.

[...]

PR #54, opened 2026-09-19
chore(infra): run the gate against its own worktree's source

Ollienel777:
Opened by the build loop.

Closes #53

## What

`scripts/gate.py` now settles which source tree it is about to exercise before
it runs a step, and prints it.

- It resolves `whiteout` for its interpreter **from a temporary directory
  outside the repository**, because `python -c` puts the working directory
  ahead of `sys.path` and a probe run at the repo root always answers "here".
- If that resolves inside this worktree, or nothing is installed at all,
  nothing changes and the existing distribution probe reports as before.
- If it resolves to **another** tree, the gate falls back to a per-worktree
  `.venv/` — created once with `--system-site-packages --without-pip`, then
  `pip install -e . --no-deps --no-build-isolation` — and every later step runs
  through that interpreter. A `.venv/` that already resolves here is reused as
  it stands, so pip runs once per worktree and not once per gate run.
- If it still does not resolve here, the gate returns
  `wrong source tree: <origin> is outside <worktree>` from the `install` step
  and exits non-zero.
- `_run` turns an `OSError` from `subprocess.run` (an unlaunchable or
  half-written `.venv/`) into exit code 127 with a printed reason, instead of a
  traceback out of the middle of a step.

`.venv` is the fallback's location because it is already in `.gitignore`, in
`[tool.ruff] exclude`, in `[tool.mypy] exclude` and in `norecursedirs`, so no
config change was needed and nothing the gate lints, typechecks, tests or
packages can pick it up.

No change to the eight steps, their order or their commands. `-m "not slow"`
and `--no-isolation` are untouched. `hackathon/SPEC.md` is not edited.

Eight new tests in `tests/test_gate_config.py`; the 16 inherited ones are
unchanged and pass.

## Why

`pip install -e .` writes one path record per Python environment. With six
worktrees sharing one environment the last to install wins, and the others'
tools import that tree's `whiteout`. M0's one-ticket-at-a-time rule is the only
thing hiding it; M1 has 25 tickets.

## How to verify

**The failure, reproduced first, on real worktrees at different commits.** The
ambient editable install on this machine pointed at
`.claude/worktrees/agent-a4e26c30e9b127643` (it moved twice more during the
session, to `agent-a60ba288c774589cc`, which is the race itself). With a probe
test asserting `whiteout.__file__` is under the tree that contains the test:

- In this worktree, `pytest -q -m "not slow"` — the literal SPEC §6 test
  command — imported
[...]

PR #52, opened 2026-09-19
feat(infra): episode record types and the versioned log

Ollienel777:
Closes #5

Opened by the build loop.

## What

Two new modules and their tests.

`whiteout/types.py` — frozen, slotted, JSON-serialisable dataclasses: `Pose`,
`SensorReport`, `WaypointIntent`, `WorldObservation`, `FleetIntent`,
`BeliefDigest`, `Contact`, `Truth`, `EpisodeRecord`. Each has `to_dict()` and a
strict `from_dict()` that rejects a missing field, an unknown field, a wrongly
typed value or an unknown contact state, naming the dotted path
(`record.observation.poses[0].speed`). Sequences are tuples, so records stay
immutable and hashable and round-trip equality is real rather than
list-vs-tuple luck.

Three supporting types beyond the nine the ticket names, because `SPEC.md` §5
forbids dicts crossing module boundaries and these are members of the named
records: `SensorFootprint` (`SensorReport.footprint`), `Detection`
(`SensorReport.detections[]`) and `TargetTruth` (`Truth.targets[]`).

`whiteout/log.py` — `SCHEMA_VERSION = 1`, `write_episode_log`,
`read_episode_log`, `iter_episode_log`, `validate_line` and
`validate_episode_log`. `EpisodeLogError` carries a 1-based `line` attribute
and prefixes every message with `line N:`. The writer emits UTF-8, LF, sorted
keys and compact separators, and refuses to write a record whose
`schema_version` is not this build's. The validator rejects, in order: a line
that is not valid JSON, a line whose `schema_version` is wrong, and a line
that fails record parsing; an empty log is rejected too.

`schema_version` is carried on **every line** rather than in a header record,
so a single line is self-describing and a partially written log is rejected at
the line that is wrong.

`truth` is a top-level field of `EpisodeRecord`, a sibling of `observation`.
`WorldObservation` has exactly `t`, `poses`, `reports`, and a test asserts both
that set and that the serialised observation subtree contains no `truth`
anywhere.

Nothing else changed. `whiteout/cli.py` still writes its placeholder log; no
acceptance criterion here required rewiring it, and the sim/scorer tickets own
that seam.

Two readings of `SPEC.md` §4 were not spelled out there and were taken the most
defensible way, noted here rather than stopped on: `Truth` holds a tuple of
targets rather than one target's fields inline (the scored scenario has one
moving target; a one-element tuple avoids a field-shape migration if that ever
[...]

PR #56, opened 2026-09-19
feat(transport): the Transport protocol and the kinematic stub

### LINKED ISSUE #6: feat: the Transport protocol and the kinematic stub
**Milestone:** M0 · **Demo beat:** 6 · **Size:** S

## What
`whiteout/transport/base.py`: a `Transport` `Protocol` with exactly
`connect()`, `observe() -> WorldObservation`, `command(FleetIntent)`,
`close()`. `whiteout/transport/kinematic.py` as a stub that returns static
poses. A factory reading `WHITEOUT_TRANSPORT`, defaulting to `kinematic`.

## Why
The seam is the single most important structural decision in the build
(`DECISION.md`). Establishing it before any policy code exists is what makes it
hold.

## Acceptance
- [ ] The protocol carries no belief, policy or scoring concept
- [ ] `WHITEOUT_TRANSPORT` unset selects `kinematic`
- [ ] An unknown value fails loudly at startup, naming the valid values
- [ ] `python -m whiteout.cli run --ticks 10` produces a valid episode log

## Notes
`SPEC.md` §4 "The seam", §7. Depends on #5.

Ollienel777:
Opened by the build loop.

Closes #6

## What

`whiteout/transport/`, the seam of `SPEC.md` §4:

- `base.py` — a `runtime_checkable` `Transport` `Protocol` with exactly
  `connect()`, `observe() -> WorldObservation`, `command(FleetIntent)`,
  `close()`, plus `TransportError` and `TRANSPORT_METHODS` (the surface, pinned
  by a test). A `Protocol` rather than an ABC so the `arena` adapter can wrap
  whatever the sponsor hands us without inheriting.
- `kinematic.py` — `KinematicTransport`, the M0 stub: one asset of each
  `VEHICLE_CLASSES` value, parked, no sensor reports, clock advancing
  0.5 s per `observe()`. Start positions are drawn once at `connect()` from a
  `numpy.random.Generator` derived from the episode seed, so the stub is a pure
  function of that seed. It opens no socket.
- `__init__.py` — the factory. `WHITEOUT_TRANSPORT` unset or empty selects
  `kinematic`; an unknown value raises `TransportError` naming `kinematic, sitl,
  arena`; `sitl` and `arena` are recognised as valid but refused as not built in
  this build, which is a different message from "not a transport".

`whiteout/cli.py`: `run` now builds the selected transport, drives it one tick
at a time, and writes real `EpisodeRecord`s through
`whiteout.log.write_episode_log` instead of `out.write_text(...)`. Its fourth
acceptance criterion is a **valid** episode log, and the #51 placeholder log is
not one the validator accepts, so the placeholder writer is gone. The record's
`belief_digest`, `contacts` and `truth` are still placeholders (finite zeros, no
contacts, no targets) — the belief field, the estimator and the sim are other
tickets, and the episode loop that fills them is #25. An unknown transport and a
`--ticks 0` run both exit 1 with a diagnostic rather than a traceback.

`tests/test_transport.py`: 22 tests, one group per acceptance criterion.

## Why

Demo beat 6, and the structural bet of `DECISION.md`: if the sponsor's arena
interface is not plain MAVLink over a socket, only the adapter is lost. That
holds only if nothing else crosses the seam, which is why the no-leak criterion
is tested mechanically rather than left to review.

## How to verify

- `python scripts/gate.py` — all eight steps.
- **No belief, policy or scoring concept crosses the seam**:
  `test_the_protocol_is_exactly_four_methods`,
  `test_observe_and_command_move_only_the_seam_types`,
  `test_the_transport_package_imports_no_belief_policy_or_scoring_module` and
  `test_the_transport_package_names_no_belief_policy_or_scoring_type`. The last
  two parse every module in `whiteout/transport/` with `ast` and check imports
  and *identifiers* (not strings, so the docstrings may name what they exclude)
  against `whiteout.belief/policy/score/sim/tune`, against `BeliefDigest`,
  `Contact`, `Truth`, `TargetTruth`, `EpisodeRecord`, and against the words
  belief/contact/truth/score/policy/entropy/allocat. Checked by appending
  `def belief_digest_for(...)` to `kinematic.py`: the test fails, and passes
  again when it is removed.
- **Unset selects `kinematic`**: `test_an_unset_transport_env_selects_kinematic`
  and `test_an_empty_transport_env_selects_kinematic`.
- **An unknown value fails loudly, naming the valid values**:
  `test_an_unknown_transport_env_fails_naming_every_valid_value` and
  `test_the_cli_refuses_an_unknown_transport_env` assert every one of
  `kinematic`, `sitl`, `arena` appears in the message. By hand:
  `WHITEOUT_TRANSPORT=kinemtaic python -m whiteout.cli run --ticks 10` prints
  `run: WHITEOUT_TRANSPORT='kinemtaic' is not a known transport; valid values
  are kinematic, sitl, arena` and exits 1, writing nothing.
- **`python -m whiteout.cli run --ticks 10` produces a valid episode log**:
  `test_the_run_command_line_produces_a_valid_episode_log` runs that exact
  command line as a subprocess and asserts `validate_episode_log(out) == 10`.
- **No egress**: `test_the_kinematic_transport_opens_no_socket` monkeypatches
  `socket.socket` and `socket.create_connection` to raise, then runs a full
  connect/observe/command/close.
- The existing suite is unchanged and still passes: 91 tests total, up from 69.

## Gate

Run on this head (`0927525`), `python scripts/gate.py`:

```
GATE PASS: install (0.3s)
GATE PASS: lint (0.2s)
GATE PASS: format (0.2s)
GATE PASS: typecheck (0.7s)
GATE PASS: test (2.9s)
GATE PASS: build (3.4s)
GATE PASS: smoke (0.9s)
GATE PASS: determinism (0.5s)

GATE OK: 8 steps in 9.1s
```

The determinism step reports `logs are byte-identical (370820 bytes)`.

## Follow-ups

- The episode record's `belief_digest`, `contacts` and `truth` are placeholders
  written by `cli.py`. #25 owns the real episode loop; the placeholder digest is
  a single function, `_placeholder_digest`, so that it is one deletion.
- `run` no longer records the seed or the transport name anywhere in the log —
  the placeholder header carried both as provenance, and `EpisodeRecord` has no
  field for them. Provenance for a written episode is a schema question, not
  one this ticket should answer.


Ollienel777 (commented):
cold: findings @ 0927525

Round 1. Reviewed at `0927525`, detached, five files. `/code-review high 56` returned
promptly and every file it named is in this diff; its findings are folded in below,
each re-verified by hand at the reviewed SHA before citing.

## What I ran

- `pip install -e ".[dev]"`, then `python -m pytest -q` → **91 passed** (69 at `f9a6b80`,
  +22 from `tests/test_transport.py`; no existing test file is touched by this diff).
- `python scripts/gate.py` → **GATE OK: 8 steps in 16.7s**. Determinism step reported
  byte-identical at **370820 bytes**, matching the checks run.
- Acceptance 4 run for real, then validated rather than trusted: `run --ticks 10` →
  `validate_episode_log(...)` → **10**.
- Determinism independently: seed 42 twice at 400 ticks → identical md5
  `0a52e8d2…`/370020 bytes; seed 43 → `9e531601…`/369620 bytes. Same seed identical,
  different seeds differ. Confirmed.
- No socket: `kinematic` imports only `numpy`, `whiteout.transport.base`,
  `whiteout.types`; `test_the_kinematic_transport_opens_no_socket`
  (`tests/test_transport.py:332`) monkeypatches `socket.socket` and
  `socket.create_connection` across connect/observe/command/close.
- The no-leak guard: mutation-tested, below.
- `whiteout/cli.py` grep for write paths: `write_episode_log` at line 113 is the only
  one. The `out.write_text` the placeholder used is gone. **`cmd_run` has exactly one
  write path**, and it is the reviewed writer — the #52 property holds.

## Acceptance, criterion by criterion

1. **No belief/policy/scoring concept on the protocol** — *met in the code as written.*
   `base.py:43` pins the surface to four methods; `observe`/`command` carry only
   `WorldObservation`/`FleetIntent` (`base.py:62,71`). Nothing in the package names a
   belief, policy, scoring or truth type. The *guard* on it is weaker than it reads —
   R1-02, R1-06.
2. **Unset selects `kinematic`** — met. `__init__.py:64` treats unset *and* empty as
   the default, which is right: an exported-but-blank shell variable is the common case.
3. **Unknown value fails loudly naming the valid values** — met, and better than asked:
   `__init__.py:66` and `:92` name the offender and all three of `kinematic, sitl,
   arena`, and `:95` separates "not a transport" from "not built yet". Verified via the
   CLI too (`tests/test_transport.py:194`, exit 1, no file left behind).
4. **`run --ticks 10` produces a valid episode log** — met, validated above.

## The CLI routing through `write_episode_log` — judged

**Correct, and the only defensible option.** The #51 placeholder emitted a `{"type":
"header", "schema": "placeholder"}` line that `iter_episode_log` rejects outright, so
criterion 4 could not have been met by patching the placeholder. Routing through the
reviewed writer is what makes "every path that replaces an episode log produces
something the reader accepts" true rather than asserted, and `--ticks 0` now refuses
(exit 1, no file) instead of writing a one-line log nothing can read
(`tests/test_transport.py:239`). `test_run_differs_across_seeds` and
`test_run_is_byte_identical_at_the_same_seed` (`tests/test_cli.py:39,31`) still pass,
and now for a *stronger* reason than before: the bytes differ because the drawn poses
differ, not because a seed was echoed into a header.

---

# Findings

## R1-01 · Critical · `whiteout/cli.py:92` (with `whiteout/transport/base.py:49-51`)

`base.py:49-51` states, of `TransportError`: *"Raised for an unknown or unimplemented
`WHITEOUT_TRANSPORT`, **and for a call made out of order** (`observe` before `connect`,
say). The CLI turns it into a diagnostic and a non-zero exit rather than a traceback."*

The CLI does not. The `try/except TransportError` closes at `cli.py:90`, covering only
`selected_transport_name()` and `create_transport()`. `transport.connect()` (`:92`) and
`observe()`/`command()` (`:95`, `:97`) are outside it. The docstring's second clause is
false as written.

Nothing can trigger it today — `create_transport` only ever returns `KinematicTransport`,
whose `connect()` cannot raise and whose `observe()` cannot raise after a successful
connect — so there is no live bug, and the fix is two lines. I am tiering it Critical
because the rubric puts a false claim in the diff there, and because the claim is aimed
squarely at the adapter authors who have not written their code yet. `SPEC.md` §7 makes
this the *normal* path for `sitl`: "Absent means the sitl transport refuses to start,
loudly." On the day that lands, `run` prints a traceback where `base.py` promised a
diagnostic.

Second defect on the same line, and the reason I would not fix this by widening the
docstring: `connect()` sits outside the `try/finally` as well, so a transport that fails
partway through `connect()` is never `close()`d and leaks whatever it opened. Move
`connect()` inside a `try` that wraps the whole drive, keep `close()` in `finally`, and
catch `TransportError` around the lot.

## R1-02 · Medium · `tests/test_transport.py:57-77`, `:98-122`

The guard on the most important structural rule in the build. I re-ran the author's own
mutation (appending a `belief_digest_for` function) — **caught**, by the word list. Then
I tried to slip things past it. I copied `whiteout/transport/` to a scratch directory,
mutated the copy, monkeypatched `TRANSPORT_PACKAGE` and ran **the real test functions**
(no PR file was edited). Three got through:

1. **Aliased import.** `from whiteout.types import BeliefDigest as _Digest`, plus a
   module-level `def latest_digest(self_t: float) -> _Digest: ...` that constructs one.
   Both guard tests **passed**. Cause: `_identifiers` at `:76` records
   `node.asname or node.name`, so the aliased-away original name is never seen; and the
   import test only compares *module* paths, and `whiteout.types` is legitimately
   allowed. This is the most plausible real leak — nobody writes
   `def belief_digest_for`, but people alias imports all the time. Fix: add **both**
   `node.name` and `node.asname` at `:76`.
2. **A subpackage.** I added `whiteout/transport/arena/adapter.py` containing
   `from whiteout.policy import allocate`, `from whiteout.types import Truth` and
   `def belief_score()`. Both tests **passed** — `_transport_sources()` at `:58` uses
   `glob("*.py")`, non-recursive, and reported scanning only
   `['__init__.py', 'base.py', 'kinematic.py']`. A subpackage is a plausible shape for
   the `arena` adapter, and it silently disables every rule. Fix: `rglob("*.py")`.
3. **Relative imports.** `ast.parse("from ..belief import prior_for")` gives
   `module='belief', level=2`. `:105` compares that bare string against fully-qualified
   `FORBIDDEN_IMPORTS`, so it matches nothing. Fix: resolve `level > 0` against the
   package before comparing.

Also slipping through, verified against a faithful reimplementation of the same logic
rather than the live tests: a string annotation (`-> "BeliefDigest"` is `ast.Constant`,
never walked), a keyword-argument name (`ast.keyword.arg` is not walked), `getattr(o,
"belief_digest")`, and `importlib.import_module("whiteout.belief")`. Those are harder to
do by accident; (1)–(3) are not. The re-export case (`__all__ = ["BeliefDigest"]`) is
caught, but only because the plain import next to it is.

`base.py:13-15` says the rule "is enforced by a test". With (1) and (2) open, that is
overstated by more than it should be for a claim this load-bearing.

## R1-03 · Medium · `whiteout/transport/__init__.py:95-101`

`create_transport` validates `name` twice and then discards it:
`return KinematicTransport(seed=seed)`, unconditionally. It is correct today only
because `IMPLEMENTED_TRANSPORTS == ("kinematic",)`.

The failure mode is the one this module's own docstring (`:9-13`) says the loud rejection
exists to prevent. The `sitl` ticket adds `"sitl"` to `IMPLEMENTED_TRANSPORTS` at `:49`,
forgets `:101`, and `WHITEOUT_TRANSPORT=sitl` silently returns the fake — "a demo run
that scored suspiciously well against a fleet that was never there", verbatim. No test
catches it: `test_a_valid_but_unbuilt_transport_fails_and_says_which_it_is` (`:181`)
explicitly `continue`s past implemented names, and no test asserts that an implemented
name constructs the class it names.

Fix: a `{"kinematic": KinematicTransport}` mapping, with `IMPLEMENTED_TRANSPORTS` derived
from its keys so the two cannot drift, and a test that each implemented name yields an
instance of its own class. This is the cheapest it will ever be to fix — one
implementation exists.

## R1-04 · Medium · `whiteout/transport/kinematic.py:83`

`np.random.default_rng(self._seed)` rejects negative integers, and nothing upstream
constrains the seed: `--seed` is `type=int` and `_resolve_seed` accepts any `int(raw)`
from `WHITEOUT_SEED`. Verified at this SHA:

```
$ python -m whiteout.cli run --seed -1 --ticks 3 --out <tmp>
rc=1
... ValueError: expected non-negative integer
```

An eight-frame numpy traceback. `_resolve_seed`'s docstring (`cli.py:42-43`) says it
returns `None` on bad input "so the caller diagnoses it instead of raising" — a negative
seed escapes exactly that. Reachable today, on the default transport. Reject it where the
other bad-seed case is rejected, with the same one-line diagnostic; do not normalise it
(`abs(seed)` would make two distinct seeds the same episode, which is worse).

## R1-05 · Medium · the dropped provenance header — judged, and it matters

The placeholder's header carried `seed` and `transport`; the episode log no longer
records either. The author flagged this rather than deciding it. My read: **it matters,
and it is cheapest to fix now**, but it should not block this PR.

Why it matters: `SPEC.md` §4 says the episode log "is the **only** artifact the scorer,
the viewer, the tuner and the ablation harness read". Demo beat 7 (§2) is "cached sweep
re-scored" — the tuner re-scores logs it produced earlier, across thousands of episodes
and two transports. A log that cannot say which seed and which transport produced it is
not self-describing, so that attribution has to live out-of-band in filenames or a side
manifest — and a side manifest is a second artifact the tuner reads, which §4 says there
is not. The first time a sweep directory gets half-regenerated, the side-table is what
goes wrong.

Why now: it is a `SCHEMA_VERSION` bump, and the bump costs "expect every artifact under
`artifacts/` to be regenerated" (`log.py`). At hour 7.5 that is zero artifacts —
`artifacts/` is gitignored and nothing committed depends on it. After the overnight sweep
it is the sweep.

I would not widen this PR for it. Open an issue against `whiteout/types.py` and
`whiteout/log.py` to add provenance to `EpisodeRecord` (or a typed first line, if the
reader is taught it), blocking the tuner ticket, and land it before any log is kept.

## R1-06 · Medium · `tests/test_transport.py:54`

`FORBIDDEN_WORDS = ("belief", "contact", "truth", "score", "policy", "entropy",
"allocat")` omits the four things the project is actually scored on. `def
coverage_fraction(poses)` and `REWARD_WEIGHT = 0.25` inside an adapter both pass the
guard today (verified against the same logic). Given that this list *is* the enforcement,
add at minimum `coverage`, `collaboration`, `efficien`, `tracking`, plus `reward`,
`estimat`, `digest` and `detection`. `score` catches `scorer` but not `coverage`, and
coverage is a scoring concept.

## R1-07 · Medium · `whiteout/transport/kinematic.py:79-83` and `:131-133`

`close()` clears `_connected` only. A later `connect()` therefore falls through the
idempotence guard at `:81`, redraws the poses and resets `self._t = 0.0`. So a
*reconnect* silently rewinds the world clock.

`test_connect_and_close_are_idempotent` (`:318`) asserts that a second `connect()` while
connected does not restart the clock — it does not cover connect-after-close, which is
the case that bites. The consequence is specific and bad: the episode loop (#25) or an
`arena` adapter recovering a dropped link mid-episode gets `observe()` restarting at
`t=0.0`, and `write_episode_log` then refuses the *whole* run at the end with "refusing
to write t …, which is before the previous record's …". The entire episode is lost
rather than the reconnect being diagnosed where it happened.

The stub's own behaviour is a detail; **what the seam does not say is the finding**.
`base.py:59-60` documents `connect` as idempotent and says nothing about reconnect, and
`arena`'s adapter will be written against that docstring by someone who has not read this
file. Decide it on the protocol now: either the clock survives a reconnect, or reconnect
is an explicit `TransportError`.

## R1-08 · Nit · `tests/test_transport.py:38-44`

`FORBIDDEN_IMPORTS` omits `whiteout.log` and `whiteout.cli`. Neither is a belief or
policy concept, but a transport that imported the episode-log writer would have coupled
the adapter to the artifact format, which is the same shape of mistake. `EpisodeRecord`
in `FORBIDDEN_TYPES` catches the likely spelling; the module ban is one line and covers
the rest.

---

## What is good here, and is worth saying

The seam itself is right. `Protocol` over ABC with the reasoning written down
(`base.py:24-29`) is the correct call for an adapter that will have to wrap an object the
sponsor hands us. `TRANSPORT_METHODS` pinned by a test makes widening the seam a
deliberate edit rather than a drift. Separating "not a transport" from "not built yet"
is a small thing that will save someone ten minutes at hour 25. The decision to treat an
empty `WHITEOUT_TRANSPORT` as unset is correct and easy to have missed. And the instinct
behind the mechanical no-leak test — that a reviewer cannot be relied on at hour 25 — is
exactly right; R1-02 is about making it live up to that, not about the idea.

## Verdict

Nothing here blocks on behaviour: the four acceptance criteria are met, the gate is
green, determinism holds, no socket is opened, and `cmd_run` has one write path through
the reviewed writer. R1-01 is a false claim in a docstring that adapter authors will
rely on, and R1-02 and R1-03 are the two places where this PR's own safety net has a hole
in it — all three are small, and all three get more expensive once there are three
adapters. R1-05 is a decision to record, not a change to make here.

Ollienel777:
cold: findings @ 0927525

Round 1 posted as a review. 1 critical, 6 medium, 1 nit. Gate re-run locally: GATE OK, 8 steps, determinism byte-identical at 370820 bytes. 91 tests pass. Acceptance 4 validated with `validate_episode_log` (10 records), not exit 0. Three mutations slipped past the no-leak guard (R1-02).

Ollienel777:
loop: fixer dispatched @ 0927525 — findings

Ollienel777:
fix: pushed @ 176d781

All eight findings reproduced at `0927525` before changing anything, and confirmed gone after. The three guard holes were re-tested round 1's way: the package copied to scratch, the copy mutated, `TRANSPORT_PACKAGE` monkeypatched, and the **real** guard functions run against it. No file under `whiteout/` was edited to test a guard.

- **R1-01** (Critical) — `cmd_run` now wraps `connect`, `observe` and `command` in both the `try/except TransportError` and the `try/finally`, so `base.TransportError`'s claim of a diagnostic rather than a traceback is true, and a connect that fails partway is still `close()`d. Its docstring now names what the CLI actually covers, and cites §7's connect-time refusal. Guarded by a parametrized test over a transport that refuses at each of the three calls: exit 1, no traceback, `close()` called exactly once, no file left behind. Reverting `connect()` back outside the `try` fails it.
- **R1-02** (Medium) — `_transport_sources` uses `rglob`; `_identifiers` records **both** `node.name` and `node.asname`; relative imports are resolved against the importing module's own package before comparison. Also added: keyword-argument names and string annotations. Closed: aliased import, subpackage, relative import, string annotation, keyword-arg name. **Left open, and now said so in `_identifiers`' docstring:** dynamic access (`getattr(o, "belief_digest")`, `importlib.import_module`), names assembled from string fragments, and any name sharing no root with `FORBIDDEN_WORDS`. `base.py` no longer says the rule "is enforced by a test" — it says it is guarded by a static test whose docstrings state its limits. Six new guard-on-the-guard tests; each fails when its own fix is reverted and no other one does, and one asserts a plausible clean `sitl.py` adapter still passes.
- **R1-03** (Medium) — `create_transport` dispatches through a new `TRANSPORT_FACTORIES` mapping, and `IMPLEMENTED_TRANSPORTS` is derived from its keys, so the two cannot drift. Because one implementation cannot distinguish a dispatch from a hard-coded return, the test registers a second transport and asserts `create_transport("sitl")` returns it; reverting to `return KinematicTransport(...)` fails it.
- **R1-04** (Medium) — `--seed -1` now prints `run: seed -1 is negative; episode seeds are non-negative` and exits 1. `KinematicTransport.__init__` also raises `TransportError` on a negative seed, for the callers #25 will add that do not go through argparse. Both surfaces have their own test; the CLI's pins the CLI's wording so the backstop cannot satisfy it.
- **R1-05** (Medium) — **left to #57**, per round 1's own recommendation. No provenance field added, no schema change.
- **R1-06** (Medium) — `FORBIDDEN_WORDS` gains `coverage`, `collaborat`, `efficien`, `tracking`, plus `reward`, `estimat`, `digest`, `detection` and `covered`. `def coverage_fraction` and `REWARD_WEIGHT` both die now.
- **R1-07** (Medium) — decided and written into the protocol: **a transport is single-use and reconnect is not supported.** One instance drives one episode; the clock is monotonic for the instance's life; recovering a dropped link is the adapter's business inside its own `connect`/`observe`, never the caller's. `KinematicTransport.connect()` after a `close()` of a live link raises `TransportError` instead of rewinding `t` to 0.0. `close()` on a link that was never up still leaves the transport usable, which is what the `finally` path needs. The existing idempotence test is unaffected.
- **R1-08** (nit) — taken. `FORBIDDEN_IMPORTS` bans `whiteout.log` and `whiteout.cli`.

Determinism: seed 42 / 400 ticks twice → identical md5 `0a52e8d245c874d061f3d5dff28f9e0e`, 370020 bytes, the same digest round 1 recorded. Seed 43 → `9e5316015eb20ec688ba60ae4d46a050`, 369620 bytes. Same seed byte-identical, different seeds differ.

`cmd_run` still has exactly one write path: `write_episode_log` at `cli.py:128` is the only writer in the file.

Gate, verbatim:

```
GATE PASS: install (0.3s)
GATE PASS: lint (0.2s)
GATE PASS: format (0.2s)
GATE PASS: typecheck (0.7s)
GATE PASS: test (3.9s)
GATE PASS: build (3.6s)
GATE PASS: smoke (1.0s)
GATE PASS: determinism (0.7s)
GATE OK: 8 steps in 10.6s
```

107 tests pass, up from 91; every step ran. `scripts/gate.py` and `tests/test_gate_config.py` untouched (#54), and `hackathon/SPEC.md` untouched.

Ollienel777 (commented):
semi-cold: does not close @ 176d781

Round 2. Reviewed at `176d781`, detached. One commit over round 1's `0927525`, five
files, +554/−49. Every round 1 finding reproduced-or-verified by re-running, not by
reading the fix note. `R1-05` is out of scope here — it is #57.

## What I ran

- `pip install -e ".[dev]"`, `python -m pytest -q` → **107 passed** (91 at `0927525`,
  +16, all in `tests/test_transport.py`). The diff deletes no behavioural assertion:
  every `-` line in the test diff is inside the guard helpers, `FORBIDDEN_WORDS`, or an
  assertion message. No other test file is touched.
- Full gate, verbatim:

```
GATE PASS: install (0.3s)
GATE PASS: lint (0.2s)
GATE PASS: format (0.2s)
GATE PASS: typecheck (0.7s)
GATE PASS: test (3.5s)
GATE PASS: build (3.4s)
GATE PASS: smoke (0.8s)
GATE PASS: determinism (0.5s)
GATE OK: 8 steps in 9.7s
```

- Determinism independently: seed 42 / 400 ticks twice → md5
  **`0a52e8d245c874d061f3d5dff28f9e0e`**, **370020 bytes**, both runs, which is round 1's
  digest byte for byte. Seed 43 → `9e5316015eb20ec688ba60ae4d46a050`, 369620 bytes.
- `cmd_run`'s write paths: grepped `cli.py` for `write_text|open(|.write(|json.dump`.
  Two hits, both `write_episode_log` (`whiteout/cli.py:24,128`). **Exactly one write
  path**, and it is the reviewed writer. The #52 property holds.
- Guard mutations: the package copied to a scratch directory, the copy mutated,
  `TRANSPORT_PACKAGE` repointed, and the **real** guard functions run. No file under
  `whiteout/` or `tests/` in this PR was edited. Revert-mutations for `R1-01` and
  `R1-03` were run against a scratch copy of the whole tree, same rule.

## Verdicts

**`R1-01` (Critical) — closed.** `whiteout/cli.py:103-125`: `connect()` (`:105`),
`observe()` (`:107`) and `command()` (`:109`) are all inside the outer
`try/except TransportError` and the inner `try/finally: transport.close()`.
`base.py:52-59` now says what the code does, and drops nothing — it still names the
out-of-order case and now cites §7's connect-time refusal. The parametrized test
(`tests/test_transport.py:719-736`) is real: it asserts exit 1, `"Traceback" not in
stderr`, `stderr.startswith("run: ")`, **`refusing.closes == 1`** — exactly once, not
merely at-least-once — and no file left behind, for `connect`, `observe` and `command`.
Verified by mutation: hoisting `transport.connect()` back above the `try` gives
`FAILED ...test_the_cli_diagnoses_a_transport_fault_and_still_closes[connect]` with the
raw `TransportError: endpoint is unset` escaping. The other two parameters still pass,
so the test discriminates the specific regression.

**`R1-03` (Medium) — closed.** The hard-coded return is gone:
`whiteout/transport/__init__.py:121` is `return TRANSPORT_FACTORIES[name](seed=seed)`,
and `IMPLEMENTED_TRANSPORTS` (`:60-62`) is a comprehension over `TRANSPORT_FACTORIES`
(`:54-56`), so the two cannot drift. Verified by mutation: restoring
`return KinematicTransport(seed=seed)` fails
`test_create_transport_dispatches_on_the_name_it_was_given` with
`asked for sitl, got KinematicTransport` (`tests/test_transport.py:469`). Registering a
second factory to make one implementation distinguishable from a fixed return is the
right shape of test; `test_every_implemented_transport_constructs_the_class_it_names`
(`:438`) alone would not have caught it.

**`R1-04` (Medium) — closed.** Both surfaces:
`python -m whiteout.cli run --seed -1 --ticks 3 --out …` → `run: seed -1 is negative;
episode seeds are non-negative`, rc=1, no file (`whiteout/cli.py:85-91`).
`KinematicTransport(seed=-1)` → `TransportError: kinematic transport: seed -1 is
negative` (`whiteout/transport/kinematic.py:58-63`), and `create_transport("kinematic",
seed=-5)` the same. No numpy frame on either path. The CLI test pinning
`"transport" not in stderr` (`tests/test_transport.py:660`) is what stops the backstop
silently satisfying the CLI's test, and it is the right assertion to have thought of.

**`R1-06` (Medium) — closed.** `FORBIDDEN_WORDS` (`tests/test_transport.py:65-82`) is
16 entries. Run against the real guard: `def coverage_fraction`, `REWARD_WEIGHT`,
`def tracking_error`, `def collaboration_bonus`, `def efficiency_ratio` — **all five
caught**. Round 1's two worked examples both die.

**`R1-08` (nit) — closed.** `FORBIDDEN_IMPORTS` (`:43-51`) carries `whiteout.cli` and
`whiteout.log`. Both caught when planted as `from whiteout.log import
write_episode_log` / `from whiteout.cli import cmd_run`. And the direction is right:
`cli.py:25` imports the transport package, never the reverse — nothing in
`whiteout/transport/` imports `whiteout.cli`.

**`R1-02` (Medium) — closed on all three, but see `R2-01`.** Re-ran round 1's leaks
against the real guards: **aliased import caught** (`kinematic.py names BeliefDigest`,
from `node.name` *and* `node.asname` at `:177-181`); **subpackage caught**
(`arena/adapter.py imports whiteout.policy`, from `rglob` at `:92`); **relative import
caught** (`kinematic.py imports whiteout.belief`, from the level-resolution at
`:119-127`). The two harder ones round 1 named are caught too: string annotation
(`-> "BeliefDigest"`, via `_string_annotations` at `:131-145`) and keyword-arg name
(`dict(belief_mass=1.0)`, via `:175-176`). I then tried five of my own: a two-level
subpackage `arena/x/y.py`, a three-dot relative import from a subpackage
(`from ...policy import allocate` — the level arithmetic at `:125` is correct for
nested packages, I checked it by hand as well as by running it), `import whiteout.sim`,
and an attribute chain `whiteout.policy.allocate`. **All four caught.** The fifth is
`R2-01`.

**`R1-07` (Medium) — closed, and I think the call is right.** Judged below.

## `R1-07`: single-use transports — the right call, for a reason the note understates

I was ready to push back on this, because `arena` may well be a reconnecting network
client and this decision is being made before that interface exists. Reading
`base.py:67-80` changed my mind, and the sentence that does it is `:76-80`: *recovery is
the adapter's business, inside its own `connect`/`observe`*. The decision does not
forbid reconnection. It forbids **caller-driven** reconnection, and leaves an adapter
free to drop and re-establish a link as many times as it likes inside a single
`observe()`, provided the `t` it reports never goes backwards.

That is the weakest constraint that makes the episode log writable at all. `log.py`
refuses a backwards `t`, so *something* has to guarantee monotonicity; the transport is
the only layer that knows whether a clock jump was a reconnect or a real rewind, and the
caller provably cannot tell the two apart. Putting the guarantee anywhere else means
discovering the violation at `write_episode_log` after 400 ticks, which costs the whole
episode. So a reconnecting `arena` adapter is fully expressible under this rule — it
just cannot ask its caller to drive the reconnect. If the sponsor's harness at 14:30Z
turns out to hand us an object whose reconnect is externally driven, the fixer is right
that it is a one-line change (`kinematic.py:97-101`, plus the docstring), and the
docstring is where a future reader will look.

Two things I checked rather than assumed. `close()` on a never-connected transport
leaves it usable (`kinematic.py:157-159`: `_closed` is only set when `_connected` was
true), which is exactly what the `finally` at `cli.py:121-122` needs when `connect()`
raised — so the fault path does not turn one `TransportError` into two. And
`test_connect_and_close_are_idempotent` (`:607`) still covers the pre-existing
behaviour, with `test_connect_after_close_refuses_rather_than_rewinding_the_clock`
(`:621`) covering the new one. The resulting `close()` contract *is* state-dependent —
`close(); connect()` is legal, `connect(); close(); connect()` is not — which is subtle,
but it is stated in both docstrings (`base.py:109-114`, `kinematic.py:150-156`) and the
asymmetry is load-bearing rather than accidental. I am not raising it.

## The stated limits: honest for `_identifiers`, wrong for `_imported_modules`

`_identifiers`' three stated limits (`tests/test_transport.py:157-163`) are accurate and
I could not beat them — dynamic access, string fragments, and names sharing no root with
the word list are genuinely the residue, and `base.py:13-17` downgrading "enforced by a
test" to "guarded by a static test … its own docstrings state what it does not catch" is
the honest sentence. The problem is that the second docstring does not hold up.

---

# Findings

## R2-01 · Medium · `tests/test_transport.py:119-121` (with `base.py:16-17`)

`_imported_modules` promises "**Every** module the source imports, with relative imports
resolved" (`:109`). It does not record what `from X import Y` binds — only `X`. So
`from whiteout import sim` appends `whiteout`, which matches nothing in
`FORBIDDEN_IMPORTS`. Run against the real guards, planted in a copy of `kinematic.py`:

```
CAUGHT  from whiteout import belief | kinematic.py names 'belief'
CAUGHT  from whiteout import policy | kinematic.py names 'policy'
CAUGHT  from whiteout import score  | kinematic.py names 'score'
LEAK!!  from whiteout import sim
LEAK!!  from whiteout import log
LEAK!!  from whiteout import cli
LEAK!!  from whiteout import tune
LEAK!!  from whiteout import log as _m          (and sim, tune — same)
LEAK!!  from ... import sim   (in arena/adapter.py)
```

Four of the seven entries in `FORBIDDEN_IMPORTS` are defeated by the most ordinary
import spelling in Python, in absolute and relative form, aliased or not, at any depth.
The other three survive only by accident — the word list catches the bare name
`belief`/`policy`/`score`, and `FORBIDDEN_IMPORTS` contributes nothing to those cases.

This is round 1's leak #3 again in a different costume, and it is plausible by accident
in exactly the way the guard exists for: `from whiteout import sim` then `sim.Vehicle(…)`
is how people import a sibling package, and `whiteout.log` was added to the list one
round ago *specifically* to stop an adapter coupling to the artifact format —
`from whiteout import log` walks straight past it.

I am not tiering this Critical, though `base.py:16-17` is now an overclaim by the same
rubric that made `R1-01` Critical. The reason is consequence: the four leaking modules
are `sim`, `log`, `cli` and `tune`, which are coupling mistakes, not belief/policy/
scoring concepts — no scored concept crosses the seam through this hole, because the
word list independently stops those three. It is a Medium that someone pays for at hour
25, not a false promise about behaviour an adapter author codes against.

Fix, one line, in the `node.level == 0` branch and its relative twin: append
`f"{module}.{alias.name}"` for each alias as well as `module` itself. Then `from
whiteout import sim` yields `whiteout.sim` and the existing comparison at `:209` does
the rest. Worth a probe test in the `_GUARDS` family, and worth correcting `:109`'s
"every module" to say what it actually resolves.

---

## Verdict

`R1-01`, `R1-02`, `R1-03`, `R1-04`, `R1-06`, `R1-07` and `R1-08` all **closed**, six of
them confirmed by mutation rather than by reading. Nothing regressed: 107 tests, gate
green on all 8 steps, determinism byte-identical at round 1's exact digest, one write
path, and the fixer resisted the temptation to widen the PR into #57.

`R1-07` in particular is a better decision than "decided something and wrote it down" —
it picks the constraint that makes the log writable while leaving `arena` room to
reconnect internally, which is the property that survives the 14:30Z workshop either
way.

`R2-01` is the one open item, and it is one line plus a probe. I would take it here
rather than as a follow-up, because this guard is the whole enforcement of §4 and it is
about to be pointed at two adapters nobody has written yet.

Ollienel777:
semi-cold: does not close @ 176d781

Round 2 posted as a review. All eight round 1 findings closed — R1-01, R1-03, R1-04 confirmed by revert-mutation; R1-02's three leaks plus the two harder ones re-run against the real guards and all caught; R1-06 and R1-08 probed. R1-07's single-use decision judged correct: it forbids caller-driven reconnect only, so a reconnecting arena adapter is still expressible. One new Medium, R2-01: `from whiteout import sim` (and log, cli, tune) walks past FORBIDDEN_IMPORTS, absolute or relative, aliased or not — one line to fix in `_imported_modules`. Gate: GATE OK, 8 steps in 9.7s. 107 tests. Determinism md5 0a52e8d2…, 370020 bytes, round 1's digest exactly. One write path.

Ollienel777:
loop: fixer dispatched @ 176d781 — findings

Ollienel777:
fix: pushed @ e0db8e2

`R2-01` — `_imported_modules` (`tests/test_transport.py:108-141`) now records `f"{base}.{alias.name}"` for every alias as well as the left-hand module, in the `level == 0` branch and the relative branch alike, so `from whiteout import log` yields `whiteout.log`; the docstring now says what is recorded (both `X` and `X.Y`, alias irrelevant) instead of "every module the source imports". Reproduced first against a scratch copy with the real guards: all eight round-2 leaks (`sim`, `log`, `cli`, `tune`, absolute, aliased, `from .. import log`, `from ... import sim` in a subpackage, two-level subpackage) went from LEAK to CAUGHT; all nine round-1/round-2 cases still caught, and the plausible clean `sitl.py`, `from . import base` and `from whiteout import types` all still pass. New parametrized probe `test_the_no_leak_guard_catches_a_module_imported_from_its_package` (3 ids): reverting the two added lines fails all three and nothing else (3 failed, 107 passed).

Further forms in the same class: one, and it is already fatal at the gate. `from whiteout import *` records only `whiteout`, because the names it binds are not in the source — ruff `F403` rejects it at the `lint` step (checked by planting it), so I left the guard alone and stated the limit in the docstring rather than special-casing the star. Everything else I re-ran — `import X`, `import X as Y`, `from X import Y as Z`, `from .`/`..`/`...` with and without a module, nested-package level arithmetic, imports inside functions and `try` blocks, attribute chains — resolves to a name `FORBIDDEN_IMPORTS` compares against.

Nothing else touched: one file, +51/−8, no change under `whiteout/`, none to `scripts/gate.py` or `tests/test_gate_config.py`. Determinism unchanged: seed 42 / 400 ticks → `0a52e8d245c874d061f3d5dff28f9e0e`, 370020 bytes; seed 43 → `9e5316015eb20ec688ba60ae4d46a050`. `cmd_run` still has exactly one write path (`whiteout/cli.py:24,128`, `write_episode_log`). 110 tests (107 + 3).

```
GATE PASS: install (0.3s)
GATE PASS: lint (0.2s)
GATE PASS: format (0.2s)
GATE PASS: typecheck (5.2s)
GATE PASS: test (3.6s)
GATE PASS: build (3.4s)
GATE PASS: smoke (0.9s)
GATE PASS: determinism (0.5s)
GATE OK: 8 steps in 14.3s
```

Ollienel777 (commented):
semi-cold: closes @ e0db8e2

Round 3. Reviewed at `e0db8e2`, detached. One commit over `176d781`, **`tests/test_transport.py`
only**, +51/−8 — confirmed: `git diff --stat 176d781..e0db8e2 -- whiteout/` is empty, so nothing
under `whiteout/` changed and every behavioural verdict from round 2 stands untouched. Narrow
round: is `R2-01` closed, and does a test-only change introduce anything new.

## `R2-01` (Medium) — closed

`tests/test_transport.py:129-141`. The two branches are unified: `base_name` is set from
`node.module` when `node.level == 0` (`:131-133`) and from the resolved package otherwise
(`:135-138`), then `:140` appends the base and `:141` appends `f"{base_name}.{alias.name}"` per
alias, skipping `*`. One code path, so the absolute and relative forms cannot drift apart again —
which is the right shape, given this is the third round of holes in the same helper.

Verified by running **the real guard functions** against mutated scratch copies of
`whiteout/transport/`, `TRANSPORT_PACKAGE` monkeypatched, via the module's own `_guard_failures`
(`:257`). No file in this PR was edited. **Every known leak spelling, 16 of them, is caught:**

```
CAUGHT  from whiteout import sim | log | cli | tune          -> whiteout.<mod>   (R2-01's four)
CAUGHT  from whiteout import log as _m / sim as _w / tune as _t                 (R2 aliased)
CAUGHT  from .. import log                       (kinematic.py)
CAUGHT  from ... import sim                      (arena/adapter.py)
CAUGHT  from whiteout.types import BeliefDigest as _D                           (R1 aliased)
CAUGHT  from whiteout.policy import allocate     (arena/adapter.py)             (R1 subpackage)
CAUGHT  from ..belief import prior_for                                          (R1 relative)
CAUGHT  import whiteout.sim   /   import whiteout.tune as _t
CAUGHT  from whiteout import cli                 (arena/x/y.py, two-level subpackage)
CAUGHT  from .... import sim                     (arena/x/y.py, level 4)
```

And **no false positive** on the three legitimate shapes: `from . import base`,
`from whiteout import types`, and the full plausible `sitl.py` adapter of `:389-413` — all three
produce zero failures. That matters more than usual here, because `arena` gets written against
this guard and a guard that rejects clean code gets deleted rather than satisfied.

The docstring is now accurate rather than an overclaim: `:109-121` says "every dotted name", spells
out that `from X import Y` records both `X` and `X.Y`, and states that the local binding is
irrelevant. That is what the code does.

The probe test (`:360-387`) is real, not decorative. Reverting only `:141` on a scratch copy of the
tree fails **exactly** `[from-package-import-module]`, `[aliased]` and `[relative]` and nothing
else — the three new ids discriminate the exact regression. (A fourth failure in that scratch run,
`tests/test_gate_config.py::test_backlog_draft_is_deleted`, is an artefact of running from a
`git archive` export with no `.git`; it passes in the real tree, and that file is #54's anyway.)

## My two fresh attempts

Rounds 1 and 2 each found a hole this way, so I wrote seven of my own rather than two. Five are
caught, and the two that are not cannot land.

1. **`.pyi` stub** — `_transport_sources()` (`:88`) is `rglob("*.py")`, so a stub file
   `whiteout/transport/leaky.pyi` containing `from whiteout import sim` is **not scanned**.
   Not a finding: a `.pyi` carries no executable import, nothing crosses the seam at runtime, and
   there is no stub file in the package or any reason for one in a 32-hour build. **Nit.**
2. **Over-deep relative import** — `from ...sim import Vehicle` in `kinematic.py` (level 3 from a
   2-deep package). `:136` computes `keep = -1`, `base = []`, so `base_name` collapses to `"sim"`
   and **nothing matches**. Also not a finding, because it is not runnable code: I planted it and
   got `ImportError: attempted relative import beyond top-level package` from
   `import whiteout.transport`, plus a ruff error and mypy
   `Cannot find implementation or library stub for module named ".sim"` — three gate steps. I
   checked the arithmetic for every *valid* level too: truncation only fires when
   `node.level > len(package)`, which is exactly the set of imports Python itself rejects. The
   level maths at `:136-138` is correct for `whiteout/transport/*.py` and for one- and two-level
   subpackages, verified by running both. **Nit.**

Caught, for the record: a `TYPE_CHECKING`-gated `from whiteout import score`; a function-local
`from whiteout import tune`; `from whiteout.tune import knobs as _k`; a leak planted in
`__init__.py`; and a parenthesised multi-alias `from whiteout import (types, cli)`. `ast.walk`
plus per-alias recording means placement and gating no longer matter, which is the property the
helper was missing.

On the star import: I agree with leaving it documented at `:122-124` rather than caught. I
re-derived it independently — `pyproject.toml:52` selects `"F"`, so `F403` fires, `lint` is gate
step 2, and `whiteout/transport/` is not excluded. Catching it in the guard would mean modelling
`__all__`, which is more machinery than the one rule ruff already enforces.

## Nothing else regressed

- `python -m pytest -q` → **110 passed** (107 at `176d781`, +3, all the new parametrized ids).
- **Determinism**: seed 42 / 400 ticks → md5 `0a52e8d245c874d061f3d5dff28f9e0e`, **370020 bytes**;
  seed 43 → `9e5316015eb20ec688ba60ae4d46a050`, 369620 bytes. Both match rounds 1 and 2 byte for
  byte, as a test-only commit requires.
- **`cmd_run` has exactly one write path.** Grepping `whiteout/cli.py` for
  `write_text|open(|write(|json.dump` gives two hits, both `write_episode_log`
  (`whiteout/cli.py:24,128`). #52's property holds.
- Full gate, verbatim:

```
GATE PASS: install (0.3s)
GATE PASS: lint (0.2s)
GATE PASS: format (0.2s)
GATE PASS: typecheck (0.7s)
GATE PASS: test (3.8s)
GATE PASS: build (3.5s)
GATE PASS: smoke (0.9s)
GATE PASS: determinism (0.5s)
GATE OK: 8 steps in 10.0s
```

  Typecheck back to 0.7s, so the parent's one-off 5.2s was a cold mypy cache, as suspected. CI's
  `gate` check is `completed/success` on this head.

## Verdict

`R2-01` **closed**, confirmed by mutation on all nine of its spellings plus seven more, with the
clean cases still passing. Nothing new at Critical or Medium: my two misses are a file extension
that carries no runtime import and an import Python, ruff and mypy each reject outright. The fix
is one code path rather than two, the docstring finally describes it, and the probe test fails for
the right reason and only that reason.

Three rounds on this helper is not a bad sign — the rule it guards is the one structural decision
in `SPEC.md` §4 that cannot be walked back at hour 25, and it now survives every spelling three
reviewers could invent. Good to merge.

Ollienel777:
semi-cold: closes @ e0db8e2

Round 3, review posted. `R2-01` closed: all 16 known leak spellings caught against the real guards on mutated scratch copies, the three clean shapes still pass, and reverting the added line fails exactly the three new parametrized ids. Two fresh misses found (a `.pyi` stub, an over-deep relative import) — both nits, neither can land. 110 tests, gate green on 8 steps, determinism `0a52e8d245c874d061f3d5dff28f9e0e` / 370020 bytes unchanged, one write path in `cmd_run`. No new Critical or Medium. No labels applied.

Ollienel777 (commented):
cold: findings @ e0db8e2

Round 4, the settling round. `/code-review high 56` was not used — reviewed by
hand (read every file at the reviewed SHA, then ran the checks below).
**1 Medium, 1 nit. Neither should block the merge**; R4-01 is a guard-coverage
hole that no code in this PR exercises, and is cheap as a follow-up.

## Gate

`PORT=3138 python scripts/gate.py` at `e0db8e2`, after `pip install -e ".[dev]"`:

```
GATE PASS: build (3.5s)
run: wrote 400 records to ...\artifacts\smoke.jsonl (transport=kinematic)
coverage: 0.0000
collaboration: 0.0000
efficiency: 0.0000
tracking_accuracy: 0.0000
total: 0.0000 (400 records)
GATE PASS: smoke (0.9s)
  logs are byte-identical (370820 bytes)
GATE PASS: determinism (0.5s)

GATE OK: 8 steps in 14.2s
```

Determinism cross-check against the recorded value: seed 42 / 400 ticks →
md5 `0a52e8d245c874d061f3d5dff28f9e0e`, 370020 bytes. Exact match, fourth round.

## Acceptance, criterion by criterion (issue #6)

1. **No belief/policy/scoring concept on the protocol** — met. `base.py:39`
   imports only `FleetIntent`, `WorldObservation`; `base.py:83-115` is exactly
   the four methods. Mutation-checked: monkeypatching a fifth method onto
   `Transport` makes `test_the_protocol_is_exactly_four_methods`
   (`tests/test_transport.py:204`) fail, so the surface is genuinely pinned.
2. **Unset selects `kinematic`** — met. Ran the CLI with `WHITEOUT_TRANSPORT`
   removed from the environment: `transport=kinematic`.
   (`whiteout/transport/__init__.py:77`.)
3. **Unknown value fails loudly naming the valid values** — met, including the
   two near-misses I tried beyond the tests: `WHITEOUT_TRANSPORT=KINEMATIC` and
   `=" kinematic"` both exit 1 with
   `... is not a known transport; valid values are kinematic, sitl, arena`
   (`__init__.py:78-83`). Refusing rather than normalising is right here.
4. **`run --ticks 10` produces a *valid* log** — met, validated rather than
   assumed: `validate_episode_log(out)` returns `10` on the log a real
   subprocess `python -m whiteout.cli run --ticks 10` wrote. `cmd_run`'s single
   write path (`whiteout/cli.py:128`) is `write_episode_log`, which round-trips
   every record through the reader (`whiteout/log.py:115-143`), so the reader
   cannot reject what this path produces.

## SPEC §4 / §7, field by field

- §4 "everything in / everything out": `observe -> WorldObservation`,
  `command(FleetIntent) -> None`, nothing else. Matches.
- §4 no network egress: `test_the_kinematic_transport_opens_no_socket`
  (`tests/test_transport.py:782`) patches both `socket.socket` and
  `socket.create_connection`. Matches.
- §7 three values in the spec's order: `TRANSPORT_NAMES`
  (`__init__.py:44`) is `("kinematic", "sitl", "arena")`, pinned by
  `tests/test_transport.py:456`.
- §7 "`sitl` … absence is normal and never blocks": `create_transport("sitl")`
  raises "not implemented in this build", distinct from "not a known
  transport" (`__init__.py:112-120`). Correct distinction.
- §4's diagram says "Transport (ABC)" where this is a `Protocol`. §8's M0 row
  says "the `Transport` Protocol", and `base.py:27-33` gives the reason
  (structural typing for a sponsor object we must wrap). Not a finding —
  §8 and the ticket are the binding text, and the Protocol is the right call
  four hours before the workshop.

## Prose against behaviour

This was the highest-yield check, so I read every docstring in the diff against
what the code does. **The prose is in good shape** — `TransportError`'s claim
about `cli run` wrapping selection *and* the whole drive (`base.py:56-59`)
holds against `cli.py:92-125`; `__init__.py:5-7`'s "only place that turns the
environment variable into an object" holds (`cmd_run` goes through
`selected_transport_name`); `kinematic.py:18-20`'s byte-identical claim is the
determinism result above. Two things did not hold.

### R4-01 · Medium · `tests/test_transport.py:108-124` and `:162-178`

The no-leak guard does not catch a forbidden module reached as an attribute of
the package root. `import whiteout` records only `whiteout` in
`_imported_modules`, and `_identifiers` sees the attribute names `sim`, `log`,
`cli`, `tune` — none of which are in `FORBIDDEN_WORDS`, exactly as
`tests/test_transport.py:380-382` notes. Run against the real guards on a
planted copy of the package:

```
PROBE attr-chain-to-sim   (import whiteout;  _w = whiteout.sim.World)        caught=False
PROBE attr-chain-to-log   (import whiteout;  _w = whiteout.log.write_...)    caught=False
PROBE aliased-root-attr   (import whiteout as _w;  _x = _w.tune.run)         caught=False
PROBE subpkg-of-sim       (import whiteout.sim.terrain)                      caught=True
PROBE from-sim-sub        (from whiteout.sim.terrain import Terrain)         caught=True
```

Why it is worth a Medium rather than a nit: this is not one of the routes the
guard says it does not catch. `_identifiers`' stated exclusions
(`:171-177`) are dynamic access, string-assembled names, and names sharing no
root with `FORBIDDEN_WORDS`; `_imported_modules`' stated exclusion (`:121-124`)
is the star import. A plain `import whiteout` and an attribute chain is none of
those, and `base.py:16-17`'s "its own docstrings state what it does not catch"
therefore overstates slightly. It is a working route, not a theoretical one:
`whiteout.log` is resolvable in any process that has run the CLI, and
`whiteout.sim` will be in every process once M1 lands — which is the moment six
implementers are writing adapters.

Note for whoever fixes it: adding `"whiteout"` to `FORBIDDEN_IMPORTS` would ban
the legitimate `from whiteout.types import ...` too, because the rule is
prefix-matched (`:223`). It wants an attribute-side rule — e.g. flag the
attribute names `sim`, `log`, `cli`, `tune`, `belief`, `policy`, `score` when
`whiteout` (or an alias of it) is imported bare — or, cheaper, a sentence in
both docstrings naming this route. I would take the sentence today and the
rule later.

I re-ran the known-clean cases after probing: a plausible `arena.py` that opens
a socket, re-establishes the link inside `observe` without rewinding `t`, and
has a `_reestablish` helper is **not** flagged. That is the round-2 ruling
holding up empirically — a reconnecting arena client is expressible, so the
14:30Z workshop cannot invalidate the single-use decision.

### R4-02 · Nit · `whiteout/transport/base.py:113-114`

`close()`'s "It is final: see the class docstring on reconnect." is unqualified,
and the implementation is more permissive: closing a link that was never up
leaves the transport usable. Verified —

```
t = KinematicTransport(seed=1); t.close(); t.connect(); t.observe().t  ->  0.0
```

— and that path is load-bearing for `test_connect_and_close_are_idempotent`
(`tests/test_transport.py:652-653`). The two neighbours get it exactly right:
`connect()` at `base.py:88-90` carves out "after `close()` has brought a *live*
link down", and `kinematic.py:151-156` says "Closing a link that was never up
leaves the transport usable … Closing a live one is final." Only this one
sentence drops the qualifier. One word ("Closing a live link is final") and it
matches its own paragraph. Nit, not blocking.

## What I did not re-do

Per the record, I did not re-attack the sixteen known leak spellings, re-derive
the determinism hashes beyond the one cross-check above, or re-open the
single-use and guard-limits decisions above nit.

Ollienel777:
cold: findings @ e0db8e2 — round 4 (settling). 1 Medium (R4-01: the no-leak guard misses `import whiteout` + `whiteout.sim|log|cli|tune.X`, a route none of its stated limits name), 1 nit (R4-02: `base.py:113` "It is final" unqualified vs close-before-connect). Neither blocks. Gate: GATE OK, 8 steps in 14.2s.

Ollienel777:
loop: known limitations @ e0db8e2

Recorded under the review bar merged in #59: Criticals block, Mediums are
recorded here and revisited only if time and usage allow. Round 4 said in its
own review that neither finding should block the merge.

**R4-01 · `tests/test_transport.py:108-124`, `:162-178`** — the no-leak guard
misses a forbidden module reached as an attribute of the package root.
`import whiteout` records only `whiteout`, and the identifier pass sees `sim`,
`log`, `cli`, `tune` as attribute names, none of which are in
`FORBIDDEN_WORDS`. Verified against planted copies: `whiteout.sim.World`,
`whiteout.log.write_episode_log` and `import whiteout as _w; _w.tune.run` all
pass, while `import whiteout.sim.terrain` is caught.

**Consequence, and why this one is worth returning to:** the route is **not**
among the guard's stated exclusions, so `base.py:16-17`'s claim that its
docstrings state what it does not catch **overstates** — the defect family
that produced six findings on PR #52. It is also a working route:
`whiteout.log` resolves in any process that has run the CLI.

**Note for whoever fixes it:** adding `"whiteout"` to `FORBIDDEN_IMPORTS` would
ban the legitimate `from whiteout.types import …`, because matching is
prefix-based. It wants an attribute-side rule, or an honest sentence in both
docstrings.

**R4-02 · `whiteout/transport/base.py:113-114`** — `close()`'s "It is final" is
unqualified, but closing a link that was never up leaves the transport usable,
and that path is load-bearing for `test_connect_and_close_are_idempotent`. Its
two neighbours carve out the never-up case correctly; only this sentence drops
the qualifier.

Ollienel777:
settled: @ e0db8e2

Settle bar under the review rule merged in #59, each clause at this head:

- the latest round is a **cold** round at `e0db8e2` and raised **no Critical**
- the **open set is empty** — every Critical ever raised is closed; the two
  open Mediums are recorded above and do not block
- **checks pass**: `gate: completed/success`
- **no human comment is unanswered**
- the verdict is a **cold reviewer's**

Four rounds. Round 1's Critical — `base.py` claiming the CLI turned a
`TransportError` into a diagnostic while the `try/except` closed before
`connect`/`observe`/`command`, with `connect()` outside the `try/finally` so a
partial connect was never closed — is fixed and pinned, with exactly-once
`close()` verified by mutation.

**The no-leak guard was attacked across three rounds**: round 1 got three
leaks past it, round 2 one, round 3 none of sixteen known spellings plus seven
fresh attempts. What it does **not** catch is now stated in its own docstring
rather than overclaimed, and `R4-01` above records the one route those stated
limits still miss.

**The single-use decision was ruled correct empirically**: a plausible
`arena.py` that opens a socket and re-establishes its link inside `observe`
without rewinding `t` is not flagged, so a reconnecting `arena` client is
expressible and today's workshop cannot invalidate the protocol.

All four acceptance criteria verified directly, criterion 4 by
`validate_episode_log` returning 10 against a log written by a real subprocess
run rather than by exit code.

Issue #57, opened 2026-09-19
feat: record episode provenance in the log


Ollienel777:
**Milestone:** M1 · **Demo beat:** 5 · **Size:** S

## What
Give the episode log a way to record what produced it — at minimum the seed
and the transport, and whatever else identifies a run — and bump
`SCHEMA_VERSION` for it.

## Why
PR #56 routed `cmd_run` through `whiteout.log.write_episode_log`, which it had
to in order to meet its own acceptance criterion that `run` produce a **valid**
episode log. The scaffold placeholder it replaced carried a seed-and-transport
header; `EpisodeRecord` has no equivalent field, so **an episode log no longer
records its own provenance.** The implementer flagged this rather than adding
a field unasked, which was right — it is a schema change.

Round 1 of #56 judged the loss and concluded **it matters**, with reasoning
worth keeping:

> `SPEC.md` §4 says the log is the *only* artifact the tuner reads, and beat 7
> re-scores cached sweeps. A schema bump costs zero artifacts now
> (`artifacts/` is gitignored and it is hour 7.5) and costs the overnight
> sweep later.

The strategic bet of this project is tuning a policy against a reimplemented
scorer over thousands of simulated episodes. Episodes that cannot be told
apart — which seed, which transport, which policy parameters — are episodes
the sweep cannot attribute. Discovering that after the sweep has run costs the
sweep.

## Acceptance
- [ ] An episode log records, at minimum, its seed and its transport
- [ ] `SCHEMA_VERSION` is bumped and the reader rejects the older version with
      a line-numbered error, as it does for every other version mismatch
- [ ] The round-trip, writer-refusal and validator guards in
      `tests/test_types_log.py` still hold, unmodified
- [ ] `whiteout.cli run` writes provenance without a second write path —
      `write_episode_log` stays the only route to `os.replace`

## Notes
`SPEC.md` §4 "The data model". Depends on #6.

**Take this before the tuner ticket.** Doing it now is a version bump against
zero stored artifacts. Doing it after the sweep is a migration plus a lost
corpus.

Issue #43, opened 2026-09-19
chore: the committed demo episode fixture


Ollienel777:
**Milestone:** M1 · **Demo beat:** 1 · **Size:** S

## What
Commit one real seeded episode log per side of beat 2 under `fixtures/episodes/`,
regenerable by a documented command, and load one by default when the viewer is
opened with no log.

## Why
The viewer's empty state should offer the next action, and every beat needs a
fallback that is a real run rather than invented data.

## Acceptance
- [ ] The fixture is regenerated by one documented command at a named seed
- [ ] Every number the viewer shows from it was computed by the sim
- [ ] Combined fixture size stays under 25 MB **across every committed episode
      log in `fixtures/episodes/`, including #48's SITL recording** — not just
      this ticket's two

## Notes
`DESIGN.md` "Data and copy voice". Depends on #25, #27, #29.

Issue #41, opened 2026-09-19
ui: the viewer polish pass against the rubric


Ollienel777:
**Milestone:** M4 · **Demo beat:** 1–7 · **Size:** M

## What
Capture every view at 1440×900, 1280×720 and 960px wide, audit against
`DESIGN.md` and the `ui-craft` rubric, fix, capture again. Two passes minimum.

## Acceptance
- [ ] Screenshots taken in-session are attached to the PR, before and after
- [ ] No raw hex or one-off size survives outside `tokens.css`
- [ ] Text contrast is at least AA throughout; focus is visible everywhere
- [ ] No layout shift on first paint; fonts preloaded
- [ ] None of the `DESIGN.md` "Don'ts" appear

## Notes
`ui-craft` skill, `DESIGN.md`. Depends on #20, #36, #38, #39, #40.

Issue #36, opened 2026-09-19
feat: score dials, fleet rows and the inspector


Ollienel777:
**Milestone:** M1 · **Demo beat:** 1 · **Size:** S

## What
`ScoreDial` ×4 in the topbar, `FleetRow` ×N and `Inspector` in the rail, per
`DESIGN.md` "Components". Every number in `--font-mono` with `tabular-nums`.

## Acceptance
- [ ] Each `ScoreDial` renders its value at `--t-24` mono tabular and its
      label at `--t-12` `--n-400`, fitting the 48px topbar
- [ ] Score values do not jitter horizontally as they tick
- [ ] Callsigns are `WING-1`/`QUAD-1`/`ROVR-1`/`TOWR-1` style, never "Vehicle 1"
- [ ] Inspector values share one right edge

## Notes
`DESIGN.md`. Depends on #18, #26.

Ollienel777:
loop: re-triage — rewrite for seven criteria, not four axes

The score dials render four axes — coverage, collaboration, efficiency,
tracking accuracy — which is what `SPEC.md` assumed. The real judging names
**seven**: search efficiency, coverage, detection speed, tracking duration,
accuracy, autonomy and collaboration, with **no published weights**.

We also do not hold the scorer any more; theirs is authoritative and we cannot
read it during a run. So this panel cannot show "our score" honestly.

**What it can honestly show**, and what the five-minute presentation needs, is
the fleet's own state: what has been searched, what is believed, who is
holding the target, how long the track has been held. That is legible,
defensible, and speaks to *autonomy* and *collaboration* — two criteria a
judge reads off behaviour.

Rewrite the acceptance criteria before building. Do not render numbers we
cannot stand behind.

PR #70, opened 2026-09-19
feat(viz): the viewer shell, tokens and states

### LINKED ISSUE #18: feat: the viewer shell, tokens and states
**Milestone:** M1 · **Demo beat:** 1 · **Size:** M

## What
`viz/index.html`, `viz/tokens.css`, `viz/viewer.css`, `viz/viewer.js`: the
three-region shell from `DESIGN.md` — topbar, field area, 320px rail, transport
bar — with every token defined and every empty/loading/error state present.
`whiteout.cli serve` binds `PORT` or an ephemeral port, never a fixed one.
Opens from `file://` too.

## Why
Beat 1, and it is the debugging surface every other M1 ticket is verified
through.

## Acceptance
- [ ] No raw hex value or one-off pixel size anywhere outside `tokens.css`
- [ ] The seven states in `DESIGN.md` all render
- [ ] Loading is a skeleton in the final layout's shape, never a spinner
- [ ] `PORT` is respected; no server reuse
- [ ] No build step, no `package.json`, no `node_modules`

## Notes
`DESIGN.md` throughout; `SPEC.md` §5 "Why the viewer is not a Node front end",
§6. Depends on #5.

Ollienel777:
Closes #18

Opened by the build loop.

## What

`viz/index.html`, `viz/tokens.css`, `viz/viewer.css`, `viz/viewer.js` and
`viz/icons.js`: the three-region shell from `DESIGN.md` — 48px topbar, field,
320px rail, 64px transport bar — with every token DESIGN.md states defined in
`tokens.css`, and each of the seven states rendered.

- `tokens.css` is the only file in `viz/` holding a raw value. It carries
  DESIGN.md's colour, type, space, radius, elevation, motion and breakpoint
  blocks verbatim, plus the structural sizes DESIGN.md states in prose
  (`--h-topbar`, `--w-rail`, `--h-transport`, `--h-control`, …).
- The seven states: no log loaded (explains, offers the bundled episode and a
  file picker), loading (a skeleton in the final layout's shape — field block,
  dial blocks in the topbar, card blocks in the rail; no spinner anywhere),
  malformed (names the failing line, the reader's schema version, and repeats
  the reader's own message), running, finished, zero contacts (the rail says
  what would post one), many contacts (the rail's list scrolls; the field does
  not).
- `whiteout serve` (new `whiteout/serve.py`) binds `PORT` or an ephemeral
  port, never a fixed one, and refuses a port something is already listening
  on rather than attaching to it. It hands out the repository root so the
  viewer at `/viz/` can read `/fixtures/` beside it.
- The page opens from `file://`: classic scripts, relative references, no
  module, no bundler, no `package.json`. There the bundled fetch is blocked by
  the opaque origin, so the page says so and the file picker takes over.
- `fixtures/episodes/demo.jsonl` is a real 120-tick seeded run written by
  `whiteout run --seed 7`, so the empty state's primary action has a valid log
  to load.

The field's layered content draw — terrain, belief, cuts, footprints, intent
lines, asset sprites, truth — is #20's. What this PR draws on the canvas is
the field's chrome: the extent graticule and the scale bar.

## Why

Demo beat 1, and the surface every other M1 ticket is debugged through. Until
this landed there was nowhere to look at an episode log.

## How to verify

```bash
PORT=3102 python -m whiteout.cli serve     # prints the URL; ctrl-c to stop
```

Then, in the browser:

1. **no log loaded** — the opening screen. "Load the bundled episode" is the
   only accent-filled control while nothing is loaded.
2. **loading** — throttle the network, or hold `viz/../fixtures/episodes/demo.jsonl`
   in devtools, and click "Load the bundled episode". The topbar dials, the
   rail cards and the field all become skeleton blocks.
3. **malformed** — "Open a log file…" and pick a log with a bad line. Copying
   `fixtures/episodes/demo.jsonl` and editing one line's `schema_version` to
   `2` gives `line 42: schema_version 2 is not supported, expected 1`, the
   same message `whiteout.log` prints.
4. **running / finished** — press play (or space). The run-state chip reads
   `episode running`, then `episode finished` in `--ok` at the last tick.
5. **zero contacts** — the bundled episode has none; the rail says what would
   post one.
6. **many contacts** — open a log whose records carry contacts; the list
   scrolls inside the rail and the field does not move.
7. **file://** — open `viz/index.html` straight off disk. The empty state
   renders; the picker loads a log; the bundled fetch is refused with a
   readable reason.

Tests, by acceptance criterion:

| criterion | test |
|---|---|
| no raw hex or one-off pixel outside `tokens.css` | `test_no_raw_hex_value_outside_tokens_css`, `test_no_one_off_pixel_size_outside_tokens_css`, `test_the_narrow_media_query_still_equals_the_bp_narrow_token` |
| every token defined, with DESIGN.md's value | `test_every_token_design_md_states_is_defined_with_its_stated_value` |
| the seven states | `test_the_three_overlay_states_are_present`, `test_the_empty_state_explains_and_offers_the_bundled_episode`, `test_the_malformed_state_names_the_failing_line_and_the_schema_version`, `test_the_running_and_finished_states_are_both_named`, `test_zero_contacts_says_what_would_post_one`, `test_many_contacts_scrolls_the_rail_and_not_the_field` |
| loading is a skeleton, never a spinner | `test_loading_is_a_skeleton_in_the_shape_of_the_layout_and_never_a_spinner` |
| `PORT` respected, no server reuse | `tests/test_serve.py` — `test_port_is_respected`, `test_an_unset_port_is_ephemeral_and_never_a_fixed_one`, `test_a_running_server_is_never_reused` |
| no build step, opens from `file://` | `test_there_is_no_build_step`, `test_the_page_opens_from_a_file_url` |
| the offered episode is real | `test_the_bundled_episode_is_a_real_valid_episode_log`, `test_the_viewer_reads_the_same_schema_version_as_the_python_reader` |

Each of those was mutation-checked: the criterion was broken in the source and
the named test went red.

## Gate

`python scripts/gate.py`, in this worktree:

```
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (2.8s)
GATE PASS: build (0.5s)
GATE PASS: smoke (0.3s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.1s
```

## Screens touched

The viewer is the only screen. Every state was run and screenshotted in
session at **1440×900**, **1280×720** and **960×720** (DESIGN.md sets
`--bp-narrow: 960px` and says mobile is not supported, so 390px was not a
target), plus the page opened from `file://` at 1440×900. Four capture passes;
what the first three found and this PR fixes:

- the primary transport control came out grey — `.btn-glyph` was declared
  after `.btn-primary` at equal specificity, so nothing on screen was
  accent-filled once a log was loaded
- the graticule stopped at the extent square, leaving an unruled band down
  each side of the field
- the rail's three tracks were fixed, so an empty contacts panel left a void
  down the middle of the rail
- below `--bp-narrow` the stacked panels were clipped mid-row
- the loading skeleton's field block had no height and was almost invisible
  against the page
- the focus ring is the accent and so is the primary control's fill, so the
  ring was invisible on exactly the control most likely to be focused
- the file input was `visually-hidden` but still on the tab route
- the "could not be read" case was titled "malformed", which it is not

## Follow-ups

Out of scope here, noticed while building:

1. **The four score dials and the four topbar meta slots read `—`.** The
   episode log carries no scores (#26) and no provenance — seed, transport,
   policy, terrain are not fields of schema 1, which is #57. The shell shows
   an em dash rather than a zero, because DESIGN.md's "Data and copy voice"
   says there is no path by which a value the system did not produce reaches
   the screen. #36 and #57 fill them.
2. **The fleet's callsigns are `fw-1`, `quad-1`, `rover-1`, `tower-1`.**
   DESIGN.md asks for `WING-1`, `QUAD-1`, `TOWR-1`; they come from
   `whiteout/transport/kinematic.py`'s stub `FLEET`, and `ARENA.md` says there
   is no rover at all — the real fleet is two towers, one quad, one plane. The
   viewer renders whatever the log names. Worth a ticket against the
   transport, not the viewer.
3. **Fonts are not self-hosted.** DESIGN.md asks for Inter and JetBrains Mono,
   self-hosted and preloaded. The token's fallback chain (`ui-sans-serif`,
   `ui-monospace`) is what renders today; there is no web font fetch, so there
   is no layout shift either. #41 owns "fonts preloaded".
4. **`fixtures/episodes/demo.jsonl` is a placeholder for #43.** It is a real
   seeded run, so every number on screen is one the system produced, but its
   fleet is parked and its belief digest is the CLI's placeholder. #43 should
   replace the file at the same path once the episode loop lands.

---
_Generated by [Claude Code](https://claude.ai/code/session_016wVo9wLab3XBjuDV4ME574)_


Ollienel777:
cold: findings @ ca87f29

Round 1. 0 Critical, 6 Medium, 8 Nit. Findings in the review.

Verified clean: all five acceptance criteria in #18; the seven states rendered and screenshotted at 1440x900; the gate (8/8 PASS in a fresh worktree); the PR body's mutation-check claim spot-checked on six of the named tests (all six went red); `fixtures/episodes/demo.jsonl` byte-identical to `whiteout run --seed 7 --ticks 120`; the `file://` path behaving exactly as the body describes.

---

Ollienel777 (commented):
cold: findings @ ca87f29

Fresh worktree at `ca87f29`. Gate run here, verbatim:

```
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (2.8s)
GATE PASS: build (0.5s)
GATE PASS: smoke (0.3s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.2s
```

`gate` on `ca87f29` is `completed success`; nothing red.

App run on `PORT=3106`, all seven states screenshotted at 1440x900 (plus 1280x720, 960x720 and `file://`), against `hackathon/DESIGN.md` and the `ui-craft` rubric.

---

## Acceptance (#18) — all five met

| criterion | verdict | how checked |
|---|---|---|
| no raw hex or one-off pixel outside `tokens.css` | met | a grep for `rgb(`, `hsl(` and named colours over `viz/*.css`, `*.html`, `*.js` is empty; the only non-token lengths are two `letter-spacing` values in `em` (`viz/viewer.css:82`, `:376`), which the criterion's wording does not reach. Three font weights only (400/500/600). |
| the seven states render | met | screenshotted, each one, at 1440x900 |
| loading is a skeleton, never a spinner | met | stubbed `window.fetch` to a never-resolving promise: dial blocks in the topbar, a field block, card blocks in all three rail panels, one slow opacity animation, no rotation anywhere |
| `PORT` respected, no server reuse | met | bound 3106 on demand; a second bind on a live port raises `ServeError`; `allow_reuse_address = False` at `whiteout/serve.py:85` |
| no build step | met | no `package.json`, `node_modules` or `tsconfig.json` at the root or in `viz/` |

## Claims in the diff, checked

All held. Specifically:

- `fixtures/episodes/demo.jsonl` is a real seeded run — `python -m whiteout.cli run --seed 7 --ticks 120` reproduces it **byte-identically**.
- `file://`: the bundled fetch is refused and the page says why — title reads `episode log could not be read`, detail reads `a page opened from file:// may not read ../fixtures/episodes/demo.jsonl, because the file:// origin is opaque. Open a log with the file picker, or serve the viewer with 'whiteout serve'.`
- malformed: editing line 42's `schema_version` to `2` gives exactly `line 42: schema_version 2 is not supported, expected 1`, as the body says.
- the focus-ring fix at `viz/viewer.css:34-40` works — the ring is plainly visible on the accent-filled play control.
- **mutation-check claim.** The body says each acceptance-criterion test was mutation-checked. I broke six of the named ones in source and all six went red: `test_many_contacts_scrolls_the_rail_and_not_the_field` (dropped `overflow-y: auto`), `test_the_running_and_finished_states_are_both_named` (renamed the string), `test_no_one_off_pixel_size_outside_tokens_css` (added `padding: 7px`), `test_a_running_server_is_never_reused` (`allow_reuse_address = True`), `test_loading_is_a_skeleton_...` (removed `sk-field`), `test_every_token_design_md_states_is_defined_with_its_stated_value` (`--accent: #38d8e1`), plus `test_port_is_respected` (`resolve_port` returning 0). The claim stands. Tree restored clean after each.

---

## Medium

**R1-M1 — the `1x` speed chip is painted in the accent while disabled, on the first screen a judge sees.**
`viz/viewer.css:302` (`.btn:disabled, .btn-glyph:disabled, .chip:disabled`) and `viz/viewer.css:332` (`.chip[aria-pressed="true"]`) have equal specificity (0,2,0) and the pressed rule is declared later, so it wins. Computed on the empty state: `disabled: true, color: rgb(56, 216, 224), background: rgba(56, 216, 224, 0.12), border: rgb(56, 216, 224)`, while `5x`/`20x`/`100x` correctly render as grey disabled boxes. So beat 1 has two accent things on screen — the primary button, which is right, and a disabled chip, which is not. This is the same specificity trap you already fixed for `.btn-glyph.btn-primary` at `viz/viewer.css:281-288`; `.chip:disabled` needs the same treatment (or `.chip[aria-pressed="true"]:not(:disabled)`). Also visible on the malformed state.

**R1-M2 — pressing play shifts the whole transport bar sideways.**
`.run-state` (`viz/viewer.css:600`) is inline content in a `flex: none` readout (`viz/viewer.css:588`), and its text changes length between `paused`, `episode running` and `episode finished`. `.scrub` is `flex: 1`, so it absorbs the difference. Measured at 1440x900 on `demo.jsonl`: scrub width `864.80` paused, `811.56` running, `810.52` finished. Every play/pause and the finish moment jumps the scrub track, the playhead and the mono clock about 53px. DESIGN.md's own type rule ("A score that jitters horizontally as it ticks is the single fastest way to look unfinished") is about exactly this, and the clock is the readout it jitters. A fixed min-width on `.transport-readout`, or reserving the chip's widest string, fixes it.

**R1-M3 — "Open a log file..." silently does nothing when you pick the same file twice.**
`viz/viewer.js:670-682`: the `change` handler never clears `el.fileInput.value`, so re-selecting the same path fires no `change`. Confirmed in-browser: after loading `demo.jsonl` and seeking to tick 51, picking `demo.jsonl` again leaves `state.index` at 50 and `input.value` at `C:\fakepath\demo.jsonl` — nothing reloaded, nothing on screen said so. This is the loop the viewer exists for: regenerate a log, re-open it, see the change. Set `el.fileInput.value = ""` after `readAsText`.

**R1-M4 — `GET /?log=...` serves a directory listing of the repository root instead of the viewer.**
`whiteout/serve.py:100-106` matches `self.path` against the literals `"/"` and `"/index.html"`, but `self.path` carries the query. Against the live server:

```
$ curl -s "http://127.0.0.1:3106/?log=fixtures/episodes/demo.jsonl"
<title>Directory listing for /?log=fixtures/episodes/demo.jsonl</title>
```

`viz/README.md` documents `?log=<path>` as the entry point, so this is the documented shortcut failing from the obvious URL, and it hands out a browsable repo tree instead. `do_HEAD` bypasses the redirect for the same reason (`HEAD /` gives a 200 listing, not a 302). Split on `?`, match the path part, and re-append the query to `Location`.

**R1-M5 — `test_the_root_leads_to_the_viewer` cannot fail.**
`tests/test_serve.py:131-134`. I replaced the redirect condition at `whiteout/serve.py:101` with `if False:` and the test still passed, because without the redirect `GET /` returns the directory listing, which is also `200`; and the second assertion, `VIEWER_INDEX in viewer_url(running)`, is a substring check on a format string that cannot be false. Assert on the final URL (`response.geturl()`), or on the body containing `WHITEOUT run viewer`. This is the test that would have caught R1-M4.

**R1-M6 — the inspector's last row is cut in half at the bottom of the rail.**
`viz/viewer.css:359-361`: `.panel-contacts` may take 50% of the rail and `.panel-inspector` absorbs what is left. With contacts present and a fleet row selected — the documented interaction, `viz/index.html:168` "Pick a fleet row to read its pose" — the inspector's nine rows do not fit: measured `scrollHeight 192` against `clientHeight 173`, last row bottom `839` against a box bottom of `836`. On screen `task / sweep sector 10` is sliced through the x-height with no scrollbar and no other cue, which reads as clipping rather than as scrolling. Reproduce: load a log whose records carry contacts, scrub to a tick with ~14 of them, click a fleet row. A lower contacts cap, or a `min-height` on the inspector body, closes it.

## Nit

**R1-N1** — the rail's left edge is not shared. `viz/viewer.css:407-409` gives `.fleet-btn` a `-8px` margin, `8px` padding and a `2px` transparent `border-left`, which nets `+2px`. Measured x at 1440x900: panel headings `1137`, contact ids `1137`, inspector keys `1137`, fleet class dots `1139`. DESIGN.md "Layout": "Everything in the rail shares a left edge."

**R1-N2** — selecting a fleet row throws focus away and is never announced. `viz/viewer.js:431-434` calls `render()`, which `clear()`s `el.fleetList` and so removes the element handling the click; `document.activeElement` is `BODY` afterwards, and during playback every tick does it again. Related: `aria-selected` is set on the `<li>` (`viz/viewer.js:422`), which as an implicit `listitem` does not support it — it belongs on the `[data-select]` button as `aria-pressed`.

**R1-N3** — `data-loading-source` (`viz/index.html:81`) is never bound in `wire()` and never written by any render; the loading state's source line is dead markup that always reads "reading the episode log".

**R1-N4** — `measureExtent` propagates `NaN`. `viz/viewer.js:516` does `Math.abs(Number(pose.x)) * 1.2` while `validateLine` only checks that `observation.poses` is an array (`viz/viewer.js:142`), so one pose missing `x`/`y` — a hand-edited or foreign-producer log, which is what the error state is for — makes `state.extent` `NaN`, and the graticule, axes and scale bar all silently vanish for the whole episode instead of failing loudly. `isFinite` per pose. Same root: the JS validator is a good deal weaker than the `whiteout.log.validate_line` it says it mirrors (no non-finite check, no clock check, no pose typing), so the viewer accepts logs the scorer rejects — worth knowing, given this is the surface every other M1 ticket is debugged through.

**R1-N5** — `reader.onerror` (`viz/viewer.js:677-680`) is the only failure path that neither calls `stop()` nor clears `state.records`; compare `viz/viewer.js:189` and `:221`. Play, then open a file whose read fails, and the rAF loop keeps advancing over the stale episode behind the error overlay.

**R1-N6** — `finished = ... && last > 0` (`viz/viewer.js:348`) means a one-record log can never reach state 5 of 7; the chip stays `paused`.

**R1-N7** — `inspectorCount` (`viz/viewer.js:475`) reads `asset` whenever `state.selected` is truthy, but `inspectorRows` falls back to the episode rows when the selected asset has no pose in this tick, so the panel is labelled `asset` while showing `source`/`schema`/`ticks`.

**R1-N8** — the dial arc is 1.75px, not 2px. `viz/index.html:219-221` draws into a 32-unit `viewBox` rendered at `--sz-dial: 28px`, so `stroke-width: var(--w-bar)` (2px, `viz/viewer.css:112`) scales by 0.875. DESIGN.md "Components" says "a thin 2px arc". Match the viewBox to the token, or divide it out.

## The four follow-ups, judged

1. **Score dials and topbar meta read an em dash.** Agree, out of scope. It is not one of #18's five criteria, and DESIGN.md's "Data and copy voice" ("There is no path by which a fake value reaches the screen") is a harder rule than the beat-1 sketch's "score dials at zero" — an invented zero would break it, and the scorer's numbers are not fields of schema 1. Worth saying out loud though: beat 1 currently opens with **eight em dashes across the top of the screen** (four meta slots, four dial values) over four empty rings, which is a lot of nothing in the first ten seconds. When #36/#57 land, that top strip is what makes the screen read as an instrument. No change asked for here.
2. **Callsigns `fw-1`/`quad-1`/`rover-1`/`tower-1`.** Agree, out of scope for the viewer — `viz/viewer.js:424` renders `pose.asset_id`, which is right, and the names come from `whiteout/transport/kinematic.py:38-43`, merged in #56. I confirmed `hackathon/ARENA.md:72-74`: "**There are no rovers.** ... it is **two towers, one quad, one plane**." File it against the transport. Note only that this PR's `fixtures/episodes/demo.jsonl` bakes that fleet in, so beat 1 shows a green `rover-1` in the rail until both are fixed together.
3. **Fonts not self-hosted.** Agree, out of scope for #18 — the criteria say nothing about fonts, and `tokens.css` carries DESIGN.md's `--font-sans`/`--font-mono` verbatim (the token test enforces it). The consequence to record for #41: every screenshot in this PR, and every one taken until #41 lands, is the `ui-sans-serif`/`ui-monospace` fallback, so self-hosting Inter and JetBrains Mono will re-flow the whole shell — the rail's `--w-inspector-key: 116px` and the topbar's four meta slots are the places to re-check.
4. **`demo.jsonl` overlaps #43.** Agree, and it is a genuinely good placeholder: byte-identical to `whiteout run --seed 7 --ticks 120`, so nothing on screen is invented. Fine as is.

## Design

The shape is right, and the restraint is the best thing about it: three regions, one accent, one animation, and a field that is deliberately empty because its content is #20's. `tokens.css` carrying DESIGN.md verbatim with a test to hold it there is the correct place to have spent the effort — it is what stops the next fifteen PRs drifting.

One thing I would push on. The seven states are implemented as two different mechanisms: three of them are `[data-view]` overlay sections driven by `setPhase`, and the other four (running, finished, zero contacts, many contacts) are emergent properties of render — a CSS `max-height`, a string in `renderTransport`, a `hidden` toggle. That is why their tests (`test_the_running_and_finished_states_are_both_named`, `test_many_contacts_scrolls_the_rail_and_not_the_field`) are string and regex greps rather than assertions about state, and why R1-M6 slipped through. If the phase enum grew to carry the run state (`empty | loading | error | paused | running | finished`) as `body[data-state]` already half does, all seven would be one mechanism, addressable from CSS and assertable from one test. Not a blocker for this PR, but worth doing before #20 layers the field draw on top of it.

No simpler shape suggests itself otherwise; `serve.py` as a 149-line `SimpleHTTPRequestHandler` subclass is the right amount of server.

**The single change that would most improve the demo**: R1-M2. The judge's first interaction with this screen is pressing play, and today that makes the bottom bar visibly jump.

---

Ollienel777:
loop: known limitations @ ca87f29

Six Mediums are open and recorded rather than fixed, per the review bar set on
issue #50 at 2026-09-19T12:25Z: Criticals block a settle, Mediums are recorded
on the PR so the cost is visible to whoever meets them later.

| id | file:line | consequence |
|---|---|---|
| R1-M1 | `viz/viewer.css` — `.chip:disabled` beaten by the later equal-specificity `.chip[aria-pressed="true"]` | the `1x` speed chip renders in the accent colour while disabled on beat 1, so the first screen a judge sees shows an active-looking control that does nothing |
| R1-M2 | `viz/viewer.js` — run-state chip | the chip's changing text resizes the scrub track by ~53px on every play/pause/finish (measured 864.8 / 811.6 / 810.5), so the transport bar visibly jumps during the demo |
| R1-M3 | `viz/viewer.js` — `fileInput.value` never cleared | re-picking the same file in "Open a log file…" is a silent no-op; a presenter retrying the same log sees nothing happen |
| R1-M4 | `whiteout/serve.py` — query not stripped before the redirect match | `GET /?log=…` serves a directory listing of the repo root instead of the viewer. `HEAD /` likewise. A deep link to a log shows the repository |
| R1-M5 | `tests/test_viz_shell.py` — `test_the_root_leads_to_the_viewer` | passes with the redirect deleted, proven by mutation, so the test does not defend the behaviour it names |
| R1-M6 | `viz/viewer.css` — inspector, rail bottom | the inspector's last row is cut mid-glyph once contacts are present and an asset is selected |

**M1, M2 and M6 are visible defects on beat 1**, which is the first screen a
judge sees. They do not block this PR under the current bar, but they are the
kind of thing the `ui-craft` rubric exists to catch, and #41 (the viewer polish
pass) is where they should land if they are not taken sooner.

**M4 and M5 are a pair**: the behaviour is wrong and the test that names it
does not defend it. Fixing M4 without M5 would leave the same hole.

Not filed as cards: at this scale a Medium backlog is one nobody reads, per
`docs/build/REVIEW.md`.

---

Ollienel777:
cold: findings @ ca87f29

Round 2, the settling round, in a fresh worktree at `ca87f29`. **0 new Critical**, 3 new Medium, 3 new Nit. Under the bar on #50 nothing here blocks a settle.

Gate run here, verbatim:

```
GATE PASS: install (1.0s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (2.9s)
GATE PASS: build (0.6s)
GATE PASS: smoke (0.4s)
GATE PASS: determinism (0.3s)
GATE OK: 8 steps in 5.4s
```

## Medium

**R2-M1 — the documented `?log=` deep link 404s for the path the viewer itself prints, and fixing R1-M4 will not change that.**

`viz/viewer.js:701-702` hands the query value straight to `loadUrl`, so it resolves against the *page's* directory, `/viz/`. `viz/README.md:30` says only "``?log=<path>`` loads a log on open", and the path the empty state puts on screen is `fixtures/episodes/demo.jsonl` (`viz/index.html:74`). Driven in Chromium against a live server on 3110:

| URL | result |
|---|---|
| `/viz/index.html?log=fixtures/episodes/demo.jsonl` | error state — `could not read fixtures/episodes/demo.jsonl: the server answered 404 File not found`; the fetch went to `/viz/fixtures/episodes/demo.jsonl` |
| `/viz/index.html?log=/fixtures/episodes/demo.jsonl` | ready, 120 records |
| `/viz/index.html?log=../fixtures/episodes/demo.jsonl` | ready, 120 records |

Worth pairing with R1-M4 rather than reading as a duplicate of it. R1-M4 is the server's redirect not stripping the query; fixing it alone would land `/?log=fixtures/…` on `/viz/index.html?log=fixtures/…`, which is the row above that 404s. The deep link needs both halves. The code already knows the right shape — `viz/viewer.js:39` spells the bundled episode `../fixtures/episodes/demo.jsonl`.

**R2-M2 — once a log is loaded there is no way back to the file picker.**

Both `data-act="open-file"` buttons live inside overlays: `viz/index.html:70` inside `[data-view="empty"]`, `viz/index.html:104` inside `[data-view="error"]`. `setPhase` (`viz/viewer.js:172-174`) hides every `[data-view]` that is not the current phase, so in `ready` neither is on screen. Measured at every step of a full drive: an `open-file` button is visible in `empty` and in `error`, and in none of the nine `ready` probes. The keyboard does not reach it either — `viz/viewer.js:684-689` binds space and the two arrows and nothing else. So pointing the viewer at a second log means reloading the page.

Two consequences. This is the surface every other M1 ticket is debugged through, and the loop it exists for — regenerate a log, look at it, regenerate, look again — costs a page reload every time. And the PR's own "How to verify" cannot be walked 1 → 7: step 6, "open a log whose records carry contacts", is unreachable from the state steps 4-5 leave you in. Together with R1-M3 (re-picking the same file is a silent no-op) the picker is reachable in two of seven states and silent in one of those two.

**R2-M3 — `whiteout serve` hands out the repository root with no deny list, gitignored secrets included.**

`whiteout/serve.py:116-132` roots the handler at `VIEWER_ROOT` (`whiteout/serve.py:41`, the repo root), and `_ViewerHandler` (`whiteout/serve.py:89-113`) adds nothing but a mime entry, the `/` redirect and a cache header. Dotfiles are served like anything else. Checked against the live server rather than inferred: I wrote a dummy `.env.local` into this worktree, `GET /.env.local` answered **200** with its contents, and the directory listing R1-M4 hands out at `/?log=…` names it in the index; I removed the file again and `git status` is clean. `.env.example` is 200 for the same reason. `CLAUDE.md` promises "Secrets never enter the repo. `.env.local` is gitignored" — that promise stops at git, not at this server, and `.env.local` is the file holding the live Mapbox token during the event. `.git/` is the same class: it 404s here only because a worktree's `.git` is a file, not a directory.

Not Critical, and I ran the checks before saying so. The bind is loopback-only (`whiteout/serve.py:48`); no CORS header is sent, so a cross-origin page cannot read a response body; and there is no write verb — `POST`, `PUT`, `DELETE` and `OPTIONS` all answer `501 Unsupported method`. A `translate_path` refusal for dotfiles, `.git` and `.env*` closes it in a few lines. Nothing bounds the tree in tests either: `tests/test_serve.py:119-128` asserts the server *does* hand out `/fixtures/…`, `:137-140` that a missing path 404s, and no test asserts anything is withheld.

## Nit

**R2-N1** — the PR body's "`tokens.css` is the only file in `viz/` holding a raw value" is broader than what holds. The acceptance criterion (no raw hex, no one-off pixel) does hold; that sentence does not — `viz/viewer.css:112` `rotate(-90deg)`, `:82` and `:376` `letter-spacing` in `em`, `:642`/`:644` `1ms`, `:654` `960px` (carved out, and held to the token by a test), and the dial geometry at `viz/index.html:219-221`.

**R2-N2** — `reducedMotion()` (`viz/viewer.js:90-92`) is dead. Nothing calls it; it is only hung on the global at `viz/viewer.js:710`, and a grep of the repo for `reducedMotion` and `WHITEOUT_VIEWER` outside `viewer.js` finds nothing. The reduced-motion behaviour that does exist is CSS (`viz/viewer.css:640-645`) plus the honest "no interpolation, so the reduced path is the same path" note at `viz/viewer.js:254-256`. The function is left over.

**R2-N3** — `test_port_is_respected` (`tests/test_serve.py:56-63`) binds the port `_a_free_port()` released a moment earlier. With `allow_reuse_address = False` and parallel worktrees, that race is exactly the flake this module is built to produce.

---

Ollienel777 (commented):
cold: findings @ ca87f29

Settling round. The head has not moved, so rather than re-verify the five criteria I went where round 1 was weakest: the demo path as a *sequence*, the claims round 1 did not check, the file-disclosure question it left at Medium, and the design split it declined to rule on. **0 new Critical, 3 new Medium, 3 new Nit** — all three Mediums are in the comment. Nothing here blocks a settle.

## Gate

`python scripts/gate.py` in this worktree, after `pip install -e ".[dev]"`, verbatim:

```
GATE PASS: install (1.0s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (2.9s)
GATE PASS: build (0.6s)
GATE PASS: smoke (0.4s)
GATE PASS: determinism (0.3s)
GATE OK: 8 steps in 5.4s
```

Working tree clean afterwards.

## The demo path, driven as a sequence

`PORT=3110 python -m whiteout.cli serve`, Chromium at 1440x900, four scripted drives with every console message, page error and failed request captured. Twenty-odd probes of `WHITEOUT_VIEWER.state` plus the DOM at each step. The sequence:

empty → load bundled → **play** → pause → 100x → play to the end → scrub back → play → scrub while playing → select a fleet row → space → space → ArrowRight.

It holds together. **No page error anywhere in any drive**; the only console errors are the deliberate 404 in the missing-log probe, a `/favicon.ico` 404, and the `file://` CORS refusal the error state is written for. Every transition leaves a coherent screen: pausing mid-play parks on the tick it reached, scrubbing during playback stops the clock first (`viz/viewer.js:665-668`), the finish moment stops the rAF loop and flips the play control's accessible name to `replay from the first tick` (`viz/viewer.js:362`), pressing play at the last tick restarts from 0 (`viz/viewer.js:235`), and `episode finished` renders in `--ok` as the body claims (`viz/viewer.css:608`, seen on screen).

Transitions round 1 could not have checked state-by-state, all clean:

- **empty → loading → ready**, with `fetch` delayed 2s so the skeleton is a real intermediate rather than a stub: 4 of 4 `[data-skeleton]` blocks visible, `#dial-row` hidden, all three real panel bodies hidden, one overlay (`loading`), **0 rotating animations**; then on `ready`, 0 skeletons visible, dial row back, all three bodies back, no overlay, transport enabled. Nothing is left behind by the transition.
- **error → ready and back**: a bad log through the picker, then the error overlay's "Load the bundled episode" recovers to 120 records; a good log after a bad one clears the overlay and repopulates the rail.
- **degradations**: HTTP 500 → `the server answered 500 Internal Server Error`; aborted request → `Failed to fetch`; an empty file → `line 1: the episode log is empty`; `?log=` pointing at a directory → the listing is read as a log and fails honestly at `line 1: not valid JSON`. All four land in the error state with the transport disabled and `records` emptied.
- **many contacts, while playing**: 14 contacts, panel `scrollHeight 464` against `clientHeight 361`, the field does not scroll and the document has no horizontal overflow at 1440.
- **`file://`**: empty state renders, the bundled fetch is refused with the message the body quotes, and the picker then loads a 120-record log.

## Claims checked

- `episode finished` in `--ok` — holds (`viz/viewer.css:608`).
- the malformed message and the `file://` message — reproduced verbatim, matching the body.
- "no build step … classic scripts, no module" — holds; `viz/icons.js` carries exactly the three glyph paths DESIGN.md allows.
- "Nothing here imports from `viz/`" (`whiteout/serve.py:19`) — holds.
- meta slots and dial values as em dashes, never an invented zero — holds; every number on screen comes from the record (`tick`, `t`, `ticks`, `belief grid/mass/covered`, contact confidences, the energy meter, which stays hidden until an asset has spent something).
- the one claim that overstates is in the PR body, not the code — R2-N1.

## Security and data loss

Nine traversal encodings against the live server — `/../etc/passwd`, `/..%2f..%2f..%2fetc/passwd`, `/%2e%2e/%2e%2e/…`, `//etc/passwd`, `/....//…`, `/viz/../../…`, the backslash-encoded form, a `%00` suffix — **all 404**. `SimpleHTTPRequestHandler.translate_path` normalises the path, so there is **no traversal out of the served root**; I can say that having run it. No write verb exists: `POST`, `PUT`, `DELETE`, `OPTIONS` all answer 501, and the viewer never writes, so there is no data-loss path.

What *is* there is disclosure **inside** the root, which is R2-M3: the root is the repository, dotfiles included, and a `.env.local` placed there is served 200 and listed. Loopback-only, no CORS headers, so it is a Medium and not a hole someone else can reach — but it is the file the harness promises never leaves the machine.

## The design split, ruled on

Round 1 asked whether the three `[data-view]` overlays versus four emergent render properties is a Critical structural problem before #20. **It is not, and collapsing the seven into one enum would be the wrong fix.**

The seven are not one dimension. `empty | loading | error | ready` is a genuine phase enum — mutually exclusive page modes that decide whether there is data at all — and it is already carried on `body[data-state]` (`viz/viewer.js:169`). Running and finished are two booleans over `ready`, recomputed per tick from `(playing, index, records.length)`. Zero and many contacts are the cardinality of one panel's data, and they are **orthogonal** to the other two: a log can be finished with many contacts, or running with none. A single enum would have to enumerate 4 x 2 x 3 to say the same thing, which is a modelling error, not a tidy-up.

The split is also smaller than it looked from the outside: the run state is already an addressable attribute, `[data-run-state]` on the chip (`viz/viewer.js:375`, `viz/index.html:202`), which `viz/viewer.css:608` styles. The only defensible move in that direction is mirroring it onto `body` so CSS can reach it from anywhere — cheap, and worth doing when #20 needs it, not before.

Where round 1's instinct was right is the *testing*, not the modelling. `test_the_running_and_finished_states_are_both_named` (`tests/test_viz_shell.py:152-157`) and `test_many_contacts_scrolls_the_rail_and_not_the_field` (`:168-175`) are greps because the four emergent states need a rendered page, not because the mechanism is wrong — and the file says so itself at `tests/test_viz_shell.py:12-14` ("What a browser is needed for … is issue #44's standing job"). The fix for R1-M6 slipping through is #44 arriving, not a refactor of the state machine. I would not hold #20 for this.

## Round 1's Mediums

I re-checked all six at this head and **every one is still exactly a Medium** — none of them was hiding a Critical underneath. R1-M1 is a disabled chip painted in the accent, visible on beat 1 and cosmetic. R1-M2 is the ~53px scrub resize, which I measured again through the sequence (`864.8` paused, `811.6` running, `810.5` finished) — it is the ugliest thing in the demo and still only a Medium. R1-M3 is silent, not destructive; R2-M2 explains why it bites harder than it reads. R1-M4 discloses the tree but cannot escape it, which is the traversal question above, answered. R1-M5 is a vacuous test, not a wrong behaviour. R1-M6 clips a row, and the rail scrolls rather than losing data. The eight nits stand as written; I did not re-raise any of them.

The shell is sound: I drove every beat-1 gesture I could think of and could not get it into a wrong or stuck state, and the reader refuses bad input in the reader's own words every time.

---

Ollienel777:
loop: known limitations @ ca87f29 (round 2)

Three further Mediums, recorded rather than fixed, per the bar on #50.
These are in addition to R1-M1..R1-M6 recorded earlier at this head.

| id | file:line | consequence |
|---|---|---|
| R2-M1 | `viz/viewer.js:701-702` | `?log=` resolves against `/viz/`, so `?log=fixtures/episodes/demo.jsonl` — **the exact path the empty state prints at `viz/index.html:74`** — 404s into the error state. Only `/fixtures/…` or `../fixtures/…` work. Fixing R1-M4 alone does not fix this; it would land the query on the row that 404s. |
| R2-M2 | `viz/index.html:70`, `:104` | both `open-file` buttons live inside the `empty` and `error` overlays, which `setPhase` hides in `ready`, and there is no keyboard path. **Loading a second log needs a page reload**, and the PR's own "How to verify" cannot be walked 1 to 7. |
| R2-M3 | `whiteout/serve.py:41,116-132` | the server hands out the repo root with no deny list. Verified live by the round: a dummy `.env.local` returned **200** and appeared in the directory listing (created and removed by the round; tree left clean). Medium rather than Critical because the bind is loopback-only, no CORS headers are sent, and POST/PUT/DELETE/OPTIONS all answer 501. |

**R2-M1 and R2-M2 are demo-path defects**, and R2-M3 plus R1-M4 are the same
root cause: the viewer's server has no notion of what it is allowed to serve.
They belong together in #41, or in a card of their own if #41 is cut.

The round found **no new Critical**. Its Critical hunt was run, not inferred:
nine traversal encodings all 404, no write verb exists, and a scripted
Chromium drive of beat 1 as a full sequence — empty, load, play, pause, 100x,
finish, scrub, play, scrub-while-playing, select, space, arrows — produced
zero page errors and no stuck or wrong state. It also re-checked all six
Mediums recorded earlier and confirmed none concealed a Critical.

It ruled on the design question round 1 deferred: the two-mechanism split for
the seven states is **not** a structural problem, and collapsing them into one
enum would be a modelling error, since contact cardinality is orthogonal to
run state. #20 should not be held for it.

---

Ollienel777:
settled: @ ca87f29

---

PR #60, opened 2026-09-19
docs(hackathon): record ArcticSim ground truth from the API workshop

Ollienel777:
`hackathon/ARENA.md` — what Dominion Dynamics actually built, from their
`ArcticSim` deck and the answers given at the 14:30Z API workshop.

**This supersedes every `assumption` in `SPEC.md` and `DECISION.md` about this
track.** Those were written when no public source about WHITEOUT existed — the
red team verified that three separate times. Most of them are now wrong.

## The four answers

- **They provide the sim.** `arctic-sim`, Docker, locally or on a cloud
  instance they issue over WireGuard.
- **The target is not adversarial** — a boat, no AIS, random spawn, random
  path. One target, no decoys.
- **Comms are free**, no bandwidth model — but the mechanism between assets is
  ours to build.
- **Judged Sunday**, one run, no leaderboard.

## What the deck adds

- **Bellot Strait, 25 km × 2 km.** A narrow water channel, not open terrain.
- **Four assets, not eleven**: two ArduPilot AntennaTracker towers, one
  Copter, one Plane. **No rovers.**
- **Detection is computer vision.** Every asset has a camera with a published
  FOV and resolution (quad 114.6°/640×480, plane 69°/640×360, towers
  60°/640×360). Finding the boat means finding it in a frame.
- **Submission is an HTTP tracks API**, not a file: `POST /api/tracks` with
  `{name, lat, lon}`, updated by name with `heading` and `speed`.
- **Seven scored criteria, not four** — search efficiency, coverage,
  **detection speed**, **tracking duration**, accuracy, **autonomy** and
  collaboration.
- **There is a five-minute presentation**, and DD named what they weigh in it:
  how you thought about the problem, how you decided to implement it, autonomy
  over manual control, and how the *robots* team up.

## Two things this breaks

**`SPEC.md`'s simulator is dead work.** Terrain generation, vehicle
kinematics, the sensor model, target movers, the episode loop and the SITL
acquisition tickets are all supplied by `arctic-sim`.

**`DECISION.md`'s strategic bet needs re-deciding.** It bets on reimplementing
the scorer and tuning over thousands of fast episodes. There is no
leaderboard, the run is judged once, the axes are seven and unweighted, and
their sim is a heavyweight Docker stack that will not run at 100× real time.

And `DECISION.md`'s repeated framing — *"scored, not judged; won by a number,
not a performance"* — is **wrong**. Half the judging is a presentation about
reasoning and autonomy.

## What survives, and is now central

[...]

PR #75, opened 2026-09-19
docs(hackathon): name the project Dogwatch, and re-plan SPEC §1, §4 and §10

### LINKED ISSUE #74: docs(hackathon): name the project Dogwatch
The project needs a name for the repo, the README and the five-minute presentation. Operator chose Dogwatch. SPEC section 1 also still claimed a single user with four scoring axes and a product tuned against a reimplemented scorer, both of which ARENA.md contradicts.

Ollienel777:
## Dogwatch

A *dogwatch* is a deliberately shortened watch at sea, and it exists for one
reason: **so the rotation shifts** and the same hands do not always draw the
same duty. It is handover, named — by sailors, centuries before this problem.

That is the shape of this project. Two towers see nothing until they are
pointed; a fixed-wing is fast with a narrow eye; a quadcopter is slow with a
wide one. The intelligence is not in any one of them, it is in the passing.

And it is the thesis the run has to demonstrate: Dominion Dynamics judge
**collaboration**, and defined it at the workshop as *how the robots team up
with each other*.

## Also in this PR: §1 was wrong twice

The section claimed:

- **one user** — "the sponsor's scoring harness… it has four numbers". There
  are **seven** criteria, no published weights, **and** a five-minute
  presentation judged on reasoning and autonomy. Two users, wanting different
  things.
- a product **"tuned against a reimplementation of the sponsor's own scoring
  function"**. That bet is dead: one judged run, no leaderboard, and a
  heavyweight Docker sim. The tickets that carried it are closed.

§1 now says both plainly, and marks that the rest of `SPEC.md` has **not** all
been re-planned yet, with `hackathon/ARENA.md` (PR #60) named as ground truth
where the two disagree. Leaving a stale one-liner at the top of the spec while
the backlog below it has been re-triaged is how someone builds the wrong thing
at 3am.

Harness change: waits for a human merge.

Closes #74


Ollienel777:
loop: extended — §4 and §10 re-planned in the same PR

Same file, same concern (bring `SPEC.md` into line with `hackathon/ARENA.md`),
so it lands together rather than as a stack of conflicting branches.

## §4 Architecture

**The seam diagram inverted, and that is the story.** `arena` was the
speculative third implementation against an unknown interface; it is now the
target. `sitl` was validation; it is now a **local dev harness**, because
ArcticSim's assets are stock ArduPilot. `kinematic` stays the fast fake CI
runs.

The seam earned its keep exactly as designed: the arena turned out **not** to
be a single MAVLink socket — it is MAVLink per asset **plus an HTTP tracks
API** — and the cost of that surprise was an adapter. The diagram now shows a
**third arrow**: a `Track` leaves by a second path, because the scored
artifact does not travel as a `FleetIntent`.

**Components:** `sim/` and `tune/` deleted, `score/` demoted to a development
instrument. **`vision/` and `tracks/` added** — the first is now the hard
part, the second is the only thing judges read.

**Two corrections that were load-bearing:**

- **"There are no external services at runtime" was false.** At runtime we
  speak MAVLink to four assets and POST to their API. What is still true is no
  keys, no accounts, no cards — so the section now says that instead.
- **There is no ground truth.** The old model had the sim write `truth` with a
  test stopping the coordinator importing it. ArcticSim gives us nothing of
  the kind, and the consequence is sharper than it sounds: **we cannot measure
  our own accuracy.** Which is why a confident wrong detection is worse than
  none.

**Where the hard part lives** — all three entries replaced. Was
negative-information, the flow network, and objective-is-the-scorer. Now
vision, negative-information, and pixel-to-lat/lon.

## §10 Risks

Rewritten. The old top risks were an unknown interface (answered), a sweep
that might not finish (closed) and a flow network that might eat a day
(closed). The new top two are **vision not working** and **the machine not
running ArcticSim**.

Also new: *we never see the vessel at all*, *a wrong lat/lon we cannot
detect*, *one judged run with no retries*, and *autonomy overclaimed*.

**The superseded table is kept below the live one rather than deleted.**
Several of its rows contain reasoning that is still correct even where the
conclusion changed — particularly the lawnmower-sweep warning, which survives
into the new table but has lost its arbiter.

## Still not re-planned

§2 demo script, §3 prize requirements, §5 stack, §6 gate, §7 environment,
§8 milestones, §9 cut list, §11 open questions. §1 marks the file as partially
re-planned and names `ARENA.md` as ground truth where they disagree.

PR #78, opened 2026-09-19
chore(infra): one coordinate frame — records are lat/lon, x/y is a drawing projection

Ollienel777:
Closes #73

Opened by the build loop.

## What

Geodetic WGS-84 `lat`/`lon` in degrees is now the frame of record, and
`whiteout/geo.py` is the one module that converts between it and a local
East–North metric frame. That local frame is a projection for drawing and
geometry only: it is never written to the episode log and never posted.

- **`whiteout/geo.py`** (new). `GeoPoint` (lat/lon, range-checked),
  `LocalPoint` (east/north metres, not a position), the WGS-84 radii of
  curvature, and both directions of the tangent-plane conversion.
  `enu_to_geodetic(lat, lon, east, north) -> GeoPoint` keeps the exact
  positional signature and semantics that `whiteout/vision/projection.py`
  on `feat/65-camera-projection` (#72) already has, so adopting it there is
  mostly an import (see Follow-ups). The module docstring states the
  tangent-plane error as a measured table rather than an adjective — it is
  **2.164 m at 3 km East, 24.039 m at 10 km East and 38.033 m at the strait
  corner**, and a few millimetres along the meridian, because the two axes
  are scaled independently and the cross term is dropped. So the error is on
  the strait's long axis, which is the one *accuracy* is scored on. The rule
  the docstring draws from that: **project about an origin near the point**,
  which is what the positional signature is for; do not take `ARENA_ORIGIN`
  and offset ten kilometres down the channel.
  `tests/test_geo.py::test_the_tangent_plane_error_is_the_one_the_module_documents`
  pins every row of that table to 1 mm against an exact ENU → ECEF → geodetic
  reference, so the numbers cannot drift back into prose.
- **`whiteout/types.py`.** Every record position is `lat`/`lon`:
  `Pose`, `SensorFootprint`, `Detection`, `Contact`, `TargetTruth`;
  `WaypointIntent.target_xy` becomes `target_lat`/`target_lon`;
  `BeliefDigest.peak_xy` becomes `peak_lat`/`peak_lon`. Construction
  range-checks each pair through `whiteout.geo`, so a swapped pair raises
  `RecordError` rather than serialising. `z`/`target_z` are untouched.
- **`whiteout/log.py`.** `SCHEMA_VERSION` 1 → 2, and
  `fixtures/episodes/demo.jsonl` is regenerated (`--seed 7 --ticks 120`).
- **`whiteout/transport/kinematic.py`.** The stub fleet is still scattered
[...]

Issue #68, opened 2026-09-19
feat: tower control and deliberate tower placement


Ollienel777:
**Milestone:** M1 · **Demo beat:** — · **Size:** M

## What
Drive the two towers — pan and tilt by raw servo PWM, plus `mode scan` — and
decide where they stand.

```
servo set 1 <pwm>   # pan  (yaw)
servo set 2 <pwm>   # tilt (pitch)
mode scan           # random scan
```

## Why
The towers are **free, permanent and never run out**, and on a 25 × 2 km
channel they are the natural tripwire. They are also the **one `.env` lever
Dominion Dynamics sanctioned**: aircraft start near-optimal and should be
left, but tower lat/lon may be chosen deliberately.

So placement is a real strategic decision with a cheap, known payoff, and it
speaks directly to *search efficiency* and *coverage*.

## Acceptance
- [ ] Commanded pan/tilt puts a tower camera on a named bearing, verified
      against the camera view
- [ ] Two placements are compared on coverage of the channel, with the
      reasoning written down — this is presentation material as well as a
      result
- [ ] Placement is data, not a hardcode, so it can be changed without a code
      change
- [ ] `mode scan` is available as the do-nothing-clever fallback

## Notes
`hackathon/ARENA.md` §3. Towers are ArduPilot **AntennaTracker**. Depends on
#63 and #65 for verification.

**Do not edit `.env` beyond tower lat/lon** — DD warned it breaks the sim.

Ollienel777:
loop: #14 closed into this ticket — you now own the "where to post" question

The flow-network ticket is closed. A min-cut over a single 25 × 2 km water
channel is degenerate: every cross-channel chord is a minimum cut, so ranking
them carries no signal.

**The real question survives and it is yours:** not *whether* to post a
watcher across the channel, but **where along the 25 km**, given two towers
with a 60° FOV, a target confined to water, and *search efficiency* and
*coverage* among the seven judged criteria.

That is geometry over a known channel, not graph theory, and it can be worked
out before the sim arrives. What #14 was really protecting — that guarding a
gap beats sweeping a field when the target must transit — is the argument this
ticket should make, and it is presentation material as much as a result.

Ollienel777:
[...]

Issue #64, opened 2026-09-19
feat: the tracks API client, and a stub server to build against


Ollienel777:
**Milestone:** M1 · **Demo beat:** — · **Size:** M

## What
A client for Dominion Dynamics' tracks API, and a stub server that mimics it
so everything above can be built and tested before the sim arrives.

```
POST http://<SIM-IP>:8010/api/tracks   {"name","lat","lon"}              -> created:true
POST http://<SIM-IP>:8010/api/tracks   {"name","lat","lon","heading","speed"} -> created:false
GET  http://<SIM-IP>:8010/api/tracks
```

A create returns `{"ok":true,"created":true,"uuid":"entity-…",…}`; a second
post with a **matching name** updates and returns `created:false`.

## Why
**This is the only artifact the judges read.** Every detection the fleet makes
scores nothing until it arrives here, and *detection speed*, *tracking
duration* and *accuracy* — three of the seven criteria — are all measured on
what this posts and how long it keeps posting.

It is also fully specified in `ARENA.md` §5, so it can be built now against a
stub and swapped to the real endpoint with a URL change.

## Acceptance
- [ ] Create, update-by-name and list, against the stub
- [ ] The sim's address comes from configuration, never hardcoded
- [ ] A failed post is retried and never silently dropped — a lost update is a
      lost scoring second
- [ ] Posting is non-blocking: a slow or dead endpoint must not stall the
      control loop
- [ ] The stub reproduces `created:true` then `created:false` on a name match

## Notes
`hackathon/ARENA.md` §5. Depends on nothing — build it first.

**Confirm with DD staff once a test detection has been submitted**, as their
deck instructs. Unknown and worth asking: whether `heading` and `speed` are
scored or optional, and how *accuracy* is measured.

Ollienel777:
Blocking decision settled: the pose timestamp contract is #80. Write the transport against that — `pose.t` stays the tick, fix age travels in a new optional `Pose.measured_t`, and the tick comes from our own monotonic counter rather than the sim's clock. #80 lands the suite change; this ticket should not carry it.

Ollienel777:
loop: ask DD what units the `speed` field expects — it is a scored field and
nothing in the build can catch a wrong guess

Raised by the operator while settling the belief-diffusion question on #79,
and it is the more urgent of the two.

`ARENA.md` §5's example payload carries `speed: 6.5` with **no stated units**:

    POST /api/tracks
[...]

PR #81, opened 2026-09-19
chore(transport): settle the pose timestamp contract

### LINKED ISSUE #80: chore: settle the pose timestamp contract before arena is written
Decision on the two timestamp assertions in `tests/transport_conformance.py`. Both are settled; `arena`, `#64` and `#68` should be written against what follows.

## A. `pose.t == observation.t` — keep it, and add a field

Neither of the offered options. Keeping it loses fix age irreversibly; relaxing it to `pose.t <= observation.t` preserves the information but **silently changes what `pose.t` means** for every consumer already reading it. Same field, same type, same plausible number, different meaning — and no test fails. That is precisely the error class `SPEC.md` §4 now records as the one we cannot detect, so it is the wrong shape of change to make.

Make it additive instead:

- **`Pose.t` keeps its meaning**: the tick this pose belongs to. `pose.t == observation.t` stays a strict equality, and an observation stays a coherent snapshot.
- **Add `Pose.measured_t: float | None = None`** — when a transport knows when the fix was actually taken, it says so there.
- **`None` means "this transport does not report measurement time"**, and a consumer that wants fix age must handle `None` explicitly.

Do **not** default `measured_t` to `t`. An adapter that forgets to set it would then report every fix as zero seconds old — plausible, wrong, and invisible. `None` forces the question.

The conformance suite gains one check: where `measured_t` is not `None`, it is finite and `measured_t <= t`.

This costs one optional field and breaks no existing caller. The relaxation costs an audit of every site that reads `pose.t`, and the audit is the part that silently fails.

**One addition worth making with it.** The kinematic transport should take a configurable pose age, defaulting to `None`. Otherwise the staleness-correction path exists only against the arena, where we get one judged run and no debugger — the fast fake should be able to hand the policy a two-second-old fix on demand.

## B. `later > earlier` — keep the strict increase

Keep it, and settle the reason alongside it: **`arena` drives its tick from our own monotonic counter, never from the sim's clock.** Then strictness is free, and the failure mode you describe — two polls of a drifting `SPEEDUP=1` clock returning the same instant, failing the adapter on tick one — cannot arise.

The observation that the docstring's own justifications do not support the assertion is correct, and the docstring should be fixed rather than the assertion. `base.py` says "monotonic", which admits equality; `log.py` accepts a repeated `t`. Neither requires a strict increase. The real reason is simpler and belongs in that docstring: **the tick is our decision cadence, not the world's clock**, `observe()` is defined to serve one tick and advance, and a tick that stands still is a coordinator that has stopped deciding while its calls keep returning.

Ollienel777:
Closes #80

Opened by the build loop.

## ⚠️ Schema — read this before merging, especially alongside #78

**`SCHEMA_VERSION` is bumped here, 1 → 2.** `measured_t` is a *required*
record key (`_check_keys` requires every declared field), so the record shape
changed, and `whiteout/log.py`'s own rule is "bump on any change to the record
shape".

Without the bump, a log written before this change and a log written after it
both self-describe as `schema_version: 1`. The version check passes and the
read fails as `record.observation.poses[0]: missing field(s) measured_t` —
naming a field instead of naming the build, which is the one message the
version number exists to produce. With the bump it reads
`schema_version 1 is not supported, expected 2`.

**#78 is open and also bumps 1 → 2**, for the unrelated lat/lon change. Both
branches change the same line to the same number, so git will merge them
**without a conflict** and `main` would end at version 2 having taken *two*
independent shape changes — the same ambiguity this bump exists to remove,
one level up.

**If both PRs land, the final number on `main` must be 3.** Whoever merges
second should:

1. set `SCHEMA_VERSION = 3` in `whiteout/log.py`,
2. set `var SCHEMA_VERSION = 3;` in `viz/viewer.js` (a test pins the two
   together), and
3. regenerate `fixtures/episodes/demo.jsonl`, taking #78's regeneration over
   this one rather than merging the two files.

If only one of the two lands, 2 is correct as it stands.

## What

Settles both timestamp assertions in `tests/transport_conformance.py`, as #80
decides.

**A.** `Pose.t` keeps meaning the tick a pose belongs to, and
`pose.t == observation.t` stays a strict equality. Fix age travels in a new
`Pose.measured_t: float | None = None` — `None` meaning "this transport does
not report a measurement time", never `t`. The conformance suite gains one
check, `check_a_reported_measurement_time_is_finite_and_never_after_its_tick`:
where `measured_t` is not `None`, it is finite and `measured_t <= t`.

**`measured_t` is in our tick timebase**, the same clock as `t`, and never the
far side's. `Pose` and `Transport.observe` both say so. A far-side stamp
(`time_boot_ms`, a `time_usec` epoch) is the *input* to the field, never its
value: an adapter must measure the offset between that clock and ours and
subtract it, or report `None`. An unconverted epoch is large enough that
`measured_t <= t` fails on tick one, but a vehicle that booted seconds ago
yields a small number that lands under `t`, passes every check, and reports an
age that means nothing — which is the failure class this whole decision was
shaped around, so it is closed in the contract rather than left to the
adapter.

**B.** `assert later > earlier` stays. Its docstring in
`tests/transport_conformance.py` and the `observe()` docstring in
`whiteout/transport/base.py` now carry the reason that supports it — the tick
is our decision cadence, not the world's clock; `observe()` serves one tick and
advances; a tick that stands still is a coordinator that has stopped deciding
while its calls keep returning. `base.py` also records that an adapter drives
its tick from its own monotonic counter, never from the far side's clock, which
is where an adapter author will read it.

**The pose age is reachable from an episode.** `KinematicTransport` takes
`pose_age_seconds` (default `None`), `create_transport` forwards it, and
`run` exposes it as `--pose-age SECONDS`. That is what makes #80's motive true
rather than only its wording: a stale fix now reaches an episode log, the
scorer, `replay` and the viewer, instead of only a unit test that builds a
transport by hand. The default forwards nothing, so an episode with no
`--pose-age` is byte-for-byte what it was. A transport that does not accept an
age refuses it by name — read off the signature, not off a caught `TypeError`,
so a constructor's own `TypeError` is not misreported as a missing argument.

Files: `whiteout/types.py`, `whiteout/log.py`, `whiteout/transport/base.py`,
`whiteout/transport/kinematic.py`, `whiteout/transport/__init__.py`,
`whiteout/cli.py`, `viz/viewer.js`, `tests/transport_conformance.py`,
`tests/test_transport_conformance.py`, `tests/test_transport.py`,
`tests/test_types_log.py`, `tests/test_cli.py`,
`fixtures/episodes/demo.jsonl`.

## Why

`arena`, #64 and #68 are written against this contract, and #80 wanted it
settled before any of them is. The configurable pose age exists so the
staleness-correction path can be driven against the fast offline fake, rather
than first executing during the one judged arena run.

## How to verify

    python scripts/gate.py
    python -m pytest tests/ -q -m "not slow"

Per acceptance criterion:

- **`pose.t == observation.t` stays an equality** —
  `tests/test_transport_conformance.py::test_the_suite_fails_a_transport_that_breaches_the_contract[_BackstampingTransport]`,
  a transport that stamps poses at their fix time. Every number is in the past,
  so a relaxed `pose.t <= observation.t` would certify it.
- **`measured_t` does not default to `t`** —
  `tests/test_types_log.py::test_a_pose_reports_no_measurement_time_by_default`
  and
  `tests/test_transport.py::test_the_stub_reports_no_measurement_time_unless_it_is_asked_to`.
- **The new check fires** — two planted breaches, per the suite's house style:
  `[_PrescientTransport]` (a `measured_t` one second after its tick) and
  `[_UnlockedTransport]` (`-inf`, chosen because it satisfies `<= t`, so only
  the finiteness clause catches it). `test_every_check_owns_a_planted_breach`
  keeps that coverage honest.
- **Reporting a measurement time is conforming** —
  `test_a_transport_that_reports_a_measurement_time_still_conforms` runs the
  whole suite against `KinematicTransport(pose_age_seconds=2.0)`.
- **The configurable pose age, and its reach** —
  `tests/test_transport.py::test_a_configured_pose_age_stamps_every_pose_behind_its_tick`,
  `::test_a_pose_age_of_zero_is_not_the_same_as_no_pose_age`,
  `::test_an_unusable_pose_age_is_refused_by_the_transport`,
  `::test_a_pose_age_that_is_not_a_number_is_refused_as_a_transport_error`,
  `::test_create_transport_carries_a_pose_age_through_to_the_transport`,
  `::test_create_transport_without_a_pose_age_reports_no_measurement_time`,
  `::test_create_transport_refuses_an_unusable_pose_age_by_name`,
  `::test_a_transport_that_does_not_take_a_pose_age_refuses_it_by_name`,
  `::test_a_constructors_own_type_error_is_not_reported_as_a_missing_pose_age`,
  `tests/test_cli.py::test_run_writes_a_stale_fix_into_the_log_when_asked`,
  `::test_run_without_a_pose_age_reports_no_measurement_time`,
  `::test_run_diagnoses_an_unusable_pose_age`.
- **The log round trip, the reader's strictness, and the version** —
  `tests/test_types_log.py::test_a_measurement_time_round_trips_through_the_log`,
  `::test_a_missing_measurement_time_is_a_rejected_record_not_a_silent_none`,
  `::test_a_wrongly_typed_measurement_time_is_rejected`,
  `::test_the_writer_refuses_a_non_finite_measurement_time`,
  `::test_a_log_of_the_shape_before_measured_t_is_refused_by_version`.
  `make_record` now carries a populated `measured_t` on one pose and `None` on
  the other, so every existing round-trip test exercises both arms.

`fixtures/episodes/demo.jsonl` is regenerated
(`python -m whiteout.cli run --seed 7 --ticks 120 --out
fixtures/episodes/demo.jsonl`). Against the previous commit its 120 lines
differ in `schema_version` alone; stripping `"measured_t":null,` and restoring
the version reproduces the `origin/main` file with zero differing lines.

Each criterion was confirmed to go red under a mutation, run with
`PYTHONDONTWRITEBYTECODE=1`, every `__pycache__` cleared between runs, and
`whiteout.__file__` checked to resolve to this worktree.

## Gate

    === [1/8] install ===
    GATE PASS: install (0.1s)
    === [2/8] lint ===
    GATE PASS: lint (0.0s)
    === [3/8] format ===
    GATE PASS: format (0.0s)
    === [4/8] typecheck ===
    GATE PASS: typecheck (0.2s)
    === [5/8] test ===
    GATE PASS: test (3.4s)
    === [6/8] build ===
    GATE PASS: build (0.5s)
    === [7/8] smoke ===
    GATE PASS: smoke (0.5s)
    === [8/8] determinism ===
    GATE PASS: determinism (0.2s)
    GATE OK: 8 steps in 4.9s

All eight steps ran here.

## Follow-ups

- **`hackathon/SPEC.md:201`'s data model line still lists `Pose` without
  `measured_t`.** Not edited here: `SPEC.md` is a harness path and editing it
  would make this PR need a human merge, which it currently does not. It is
  one word, and **#78 already edits `SPEC.md` and already needs a human
  merge**, so it should ride with that PR.
- **`SPEC.md` §4 does not record the "same field, same type, different
  meaning" error class**, though #80's text says it does. The `Pose` docstring
  no longer cites it and states the argument on its own terms. If §4 should
  genuinely gain that passage, that is a separate harness change for a human
  to make — worth doing, since #64, #68 and `arena` are all described as
  written against §4.
- Nothing consumes `measured_t` yet. The first consumer (the estimator's
  staleness correction) has to handle `None` explicitly; that is #80's point
  and belongs to the ticket that writes it.

---


Ollienel777:
cold: findings @ 145b830

2 Critical, 3 Medium, 2 Nit. Full findings in the review on this commit.

Both Criticals are false prose in the diff, the same species as the three found
on this suite before it merged: `whiteout/types.py:283` cites a passage
`hackathon/SPEC.md` §4 does not contain, and
`tests/transport_conformance.py:253-254` quotes `whiteout/transport/base.py` as
saying something this same diff changed it to contradict.

What I verified and found sound: the additive approach genuinely avoids the
rejected option's failure mode; `None` holds on every path I could reach;
both planted breaches fire on the assertions they target and not a neighbour;
the fixture regeneration claims are exact; twelve mutations run here, all RED.
Gate green locally (`GATE OK: 8 steps in 4.6s`), CI `gate` green on 145b830.

---

Ollienel777 (commented):
cold: findings @ 145b830

Reviewed against #80 point by point, read as the author of `arena` will read
it. All line references are at 145b830 and I opened each before citing it.

**Gate, run here at 145b830, verbatim:**

```
=== [1/8] install ===
GATE PASS: install (0.1s)
=== [2/8] lint ===
GATE PASS: lint (0.0s)
=== [3/8] format ===
GATE PASS: format (0.0s)
=== [4/8] typecheck ===
GATE PASS: typecheck (0.1s)
=== [5/8] test ===
GATE PASS: test (3.2s)
=== [6/8] build ===
GATE PASS: build (0.5s)
=== [7/8] smoke ===
GATE PASS: smoke (0.5s)
=== [8/8] determinism ===
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.6s
```

CI: `gate completed success` on 145b830. Nothing in this diff caused a check
failure.

---

## Critical

### R1-C1 — `whiteout/types.py:283` cites a passage `SPEC.md` §4 does not contain

The new `Pose` docstring closes its argument with:

> the error class ``SPEC.md`` §4 records as the one we cannot detect.

`hackathon/SPEC.md` §4 runs lines 153-216 ("Architecture" / "The seam" /
"Components" / "The data model"). It records no error class of that kind. The
nearest thing in §4 is at `hackathon/SPEC.md:211-213`, which is about ground
truth leaking into the observation — "the single easiest way to accidentally
cheat ourselves and believe a number" — a different claim about a different
mistake. `grep -n "cannot detect\|error class" hackathon/SPEC.md` returns
nothing, at 145b830 and on `origin/main`, and `SPEC.md` is untouched by this
diff (`git diff origin/main...145b830 -- hackathon/SPEC.md` is empty).

#80 asserts that §4 "now records" it. It does not, and this PR did not add it.
So the diff ships a citation that a reader who follows it finds absent — and
the sentence's whole weight rests on that citation being real, because the
argument for keeping the equality is "the spec already names this failure
class".

Fix: either add the passage to `SPEC.md` §4 in this PR and keep the citation,
or drop the citation and let the sentence stand on its own words, which it
can.

### R1-C2 — `tests/transport_conformance.py:253-254` quotes `base.py` as saying what this same diff changed it to contradict

`check_the_tick_clock_is_monotonic`'s rewritten docstring opens its case for
the strict increase with:

> The assertion below is a strict increase, and neither of the rules it sits
> next to is the reason. ``whiteout/transport/base.py`` says the clock is
> monotonic, which admits equality; …

But this PR rewrites `whiteout/transport/base.py:70-71` to:

> The tick clock :meth:`observe` returns **strictly increases** for the life
> of the instance

and `whiteout/transport/base.py:103-104` to "**``t`` is our decision cadence,
not the world's clock, and it strictly increases.**" After this diff,
`grep -n monotonic whiteout/transport/base.py` has exactly one hit, line 115,
and it is about the *adapter's own counter*, not about the clock `observe`
returns. So base.py no longer says the thing it is quoted as saying, and it
says the opposite of it.

The `log.py` half of the same sentence is true — I checked
`whiteout/log.py:280`, `record.t < previous`, a strict less-than, so a
repeated `t` is accepted. It is only the base.py clause that is false.

This matters more than a stale cross-reference because Part B of this PR is
*only* prose. The deliverable is the justification, and the justification's
first premise is falsified by the other file in the same commit. An adapter
author who follows the pointer finds base.py asserting strictness and the
docstring telling them base.py does not — and the structure of the argument
("neither of the rules it sits next to is the reason") stops holding, since
base.py is now one of the rules that *is* the reason.

Fix: reword to the true version, e.g. "base.py used to say only 'monotonic',
which admits equality; it now states the strict increase, and this is why" —
or cite only `log.py`, which does still admit equality.

---

## Medium

### R1-M1 — a new *required* log key with no `SCHEMA_VERSION` bump: two incompatible shapes both self-describe as `schema_version: 1`

Judged on its own terms, independently of #78.

`measured_t` is optional on the dataclass (`whiteout/types.py:307`) but
mandatory in a record: `_check_keys` (`whiteout/types.py:150-155`) requires
every declared field, and `_optional_float` (`whiteout/types.py:202-212`)
indexes `payload[name]` unguarded, relying on that. The record shape changed.
`whiteout/log.py:61-63` states the rule in the diff's own words: "Bump on any
change to the record shape, and expect every artifact under ``artifacts/`` to
be regenerated."

What actually happens to a v1 log, run here against this reader:

```
old line schema_version: 1
v1 line -> EpisodeLogError : line 1: record.observation.poses[0]: missing field(s) measured_t
new line schema_version: 1
```

Both files claim version 1. The reader passes the version check at
`whiteout/log.py:231-236` and then fails on a field. So the one message the
version field exists to produce — "this was written by a different build" —
is the one message that cannot come out. The inverse is worse: a pre-merge
checkout reading a post-merge log reports `unknown field(s) measured_t`, and
neither side can use the version number to tell why. The PR body's framing
("the round trip does not need it") tests the wrong thing: the round trip is
within one build; the version is for across builds, which is exactly the case
that now misreports.

Also: the PR body justifies this with "#80 said not to bump unless the round
trip requires it." #80 says nothing about `SCHEMA_VERSION` — I read it in full
and its comment thread is empty. That is a decision attributed to the operator
that they did not make.

I am not treating the #78 collision as a finding, as instructed. But #78
already bumps 1 → 2, so taking the bump here costs nothing but the merge
order, and leaving it out is the only outcome where a version 1 log is
ambiguous.

### R1-M2 — `measured_t` has no stated clock domain, and `base.py:119-122` points the adapter author at the wrong clock for it

Read as the author of `arena`: four assets, asynchronous MAVLink poses each
carrying its own timestamp, tick from our own monotonic counter.

`whiteout/transport/base.py:119-122` tells them:

> Count ticks here and stamp them; read the far side's time only where it is
> the answer to a different question.
>
> That other question is fix age, and it has its own field.

Taken literally, that says: put the far side's timestamp in `measured_t`. But
`measured_t` must be in *our* tick timebase — `whiteout/types.py:287-288`
("it is ``<= t``, and ``t - measured_t`` is the age of the fix") and the new
check at `tests/transport_conformance.py:390` both require it — and a MAVLink
timestamp is not. `GLOBAL_POSITION_INT.time_boot_ms` is milliseconds since the
autopilot booted; `time_usec` fields are microseconds, often a UNIX epoch. Our
`t` starts at 0.0 and counts our ticks.

Two outcomes, and one of them is the one #80 exists to prevent:

- A UNIX-epoch or long-uptime value is enormous relative to `t`, so
  `measured_t <= pose.t` fails on tick one — loud, fine.
- A vehicle that booted a few seconds ago gives `time_boot_ms/1000` of, say,
  1.2, landing under a tick `t` of 2.0. It passes the conformance suite and
  reports an age of 0.8 s that means nothing. Same field, same type, same
  plausible number, different meaning, and no test fails — the exact failure
  class this whole decision was shaped around, reintroduced one level down.

Neither `whiteout/types.py:285-289` nor `whiteout/transport/base.py:122-129`
says "expressed in the same clock as `t`", nor that the adapter must measure
and apply the offset between the far side's clock and ours. One sentence in
the `Pose` docstring closes it, and it is worth having before `arena` is
written rather than after.

### R1-M3 — `pose_age_seconds` cannot be reached from any episode, so the stated motive is only half met

`whiteout/transport/kinematic.py:64-68` is explicit about why the knob exists:

> the alternative is that nothing downstream ever sees a stale fix until a
> link to a real vehicle serves one, on the one run that is judged …

But `create_transport` calls `TRANSPORT_FACTORIES[name](seed=seed)` and
nothing else (`whiteout/transport/__init__.py:121`), and `whiteout/cli.py:97`
is `create_transport(name, seed=seed)` with no corresponding flag. So a
`measured_t` can only ever be non-`None` inside a directly-constructed
transport in a unit test. Confirmed on the regenerated fixture: every one of
the 120 lines carries `measured_t: null`, and the set of values across the
file is `{None}`.

That covers the policy's own unit tests, which is the literal wording of #80.
It does not cover the end of the path #80 is actually worried about — an
episode log, the scorer, `replay`, the viewer — none of which can be handed a
stale fix without editing code first. A `pose_age_seconds=` kwarg threaded
through `create_transport` and a `--pose-age` on `run` (default `None`, so
nothing regenerates) is what would make the docstring's claim true.

---

## Nit

### R1-N1 — `whiteout/transport/kinematic.py:89` coerces before it validates

`pose_age_seconds = float(pose_age_seconds)` runs at line 89, before the guard
at line 90. Observed here:

```
pose_age_seconds='soon' -> ValueError: could not convert string to float: 'soon'
pose_age_seconds='2.0'  -> accepted, measured_t=-2.0
pose_age_seconds=True   -> accepted, measured_t=-1.0
```

So a non-numeric argument escapes as a bare `ValueError`, not a
`TransportError`, and `True` is silently an age of one second. The seed guard
three lines above exists for precisely the opposite reason — "The seam's own
error type, either way" (`whiteout/transport/kinematic.py:86`). An
`isinstance(pose_age_seconds, int | float) and not isinstance(..., bool)` test
before the coercion closes it. Never blocks.

### R1-N2 — `hackathon/SPEC.md:201` still lists `Pose` without `measured_t`

```
Pose        : asset_id, cls, t, x, y, z, heading, speed, energy_used
```

§4 is the authority #64, #68 and `arena` are described as being written
against, and it is not one of the harness files `CLAUDE.md` holds for a human
merge. An implementer building from that line omits the field; the in-process
`Pose` accepts the omission and defaults to `None`, so nothing complains until
someone asks why fix age is `None` everywhere. One word.

---

## What I checked and found sound

**Acceptance against #80, point by point.** `Pose.t` keeps its meaning;
`pose.t == observation.t` is still an equality at
`tests/transport_conformance.py:335`; `measured_t: float | None = None` is at
`whiteout/types.py:307`; the one new check is
`check_a_reported_measurement_time_is_finite_and_never_after_its_tick` at
`tests/transport_conformance.py:350`, registered at line 572; `assert later >
earlier` survives at line 280; `pose_age_seconds` defaults to `None` at
`whiteout/transport/kinematic.py:80`. The additive approach does avoid the
rejected option's failure mode: a grep for `pose.t` across `whiteout/` and
`scripts/` finds no product consumer of it at all beyond docstrings, and no
code anywhere uses `astuple` or positional unpacking on a `Pose`, so appending
a defaulted field changes no existing reader.

**The `None` discipline.** I could not find a path that produces `0.0` or `t`
by accident. The dataclass default is `None` and a forgetful constructor gets
`None` (mutation 1 below). `to_dict` emits `"measured_t": null` and
`_optional_float` (`whiteout/types.py:202`) maps `null` back to `None` rather
than to a number, while `_check_keys` still rejects an *absent* key — the
distinction the field exists for. The kinematic stub stamps `None` unless
asked. `pose_age_seconds=0.0` does give `measured_t == t`, and at tick zero
that is literally `0.0` — but that is an explicit claim by a caller who typed
it, which is what `test_a_pose_age_of_zero_is_not_the_same_as_no_pose_age`
asserts, not a default. The one leak is in-process only and not new to this
field: `Pose(measured_t=False)` coerces to `0.0` via `_as_floats`, the same
laxity every other float field on the record already has; JSON cannot reach
it, because `_float` rejects a bool.

**Both planted breaches fire on the assertion they target.** I ran each check
directly and read the failing frame:

```
_BackstampingTransport: AssertionError at transport_conformance.py:335
    assert pose.t == observation.t
_PrescientTransport:    AssertionError at transport_conformance.py:390
    assert measured_t <= pose.t
_UnlockedTransport:     AssertionError at transport_conformance.py:387
    assert math.isfinite(measured_t)
```

Not a neighbour in any of the three. `-inf` is the right choice for the
finiteness clause: it satisfies `<= t`, so nothing else catches it.

**Mutations — twelve run here, all RED.** `PYTHONDONTWRITEBYTECODE=1`, every
`__pycache__` removed before and after each run, `-p no:cacheprovider`, and
`whiteout.__file__` confirmed to resolve to this worktree first. One mutation
I tried initially came back GREEN and was my mistake, not a gap: tolerating a
missing key at the `Pose.from_dict` call site is dead code, because
`_check_keys` rejects it first. Mutating `_check_keys` itself — the author's
actual claim — is RED.

| mutation | result | caught by |
|---|---|---|
| `measured_t` defaults to `0.0` | RED | `test_a_pose_reports_no_measurement_time_by_default` |
| `pose.t == observation.t` relaxed to `<=` | RED | `[_BackstampingTransport]` |
| `later > earlier` relaxed to `>=` | RED | `[_RewindingTransport]` |
| new check drops `measured_t <= t` | RED | `[_PrescientTransport]` |
| new check drops the finiteness clause | RED | `[_UnlockedTransport]` |
| stub's pose age defaults to `0.0` | RED | `test_the_stub_reports_no_measurement_time_unless_it_is_asked_to` |
| new check unregistered from `CONFORMANCE_CHECKS` | RED | `test_every_check_owns_a_planted_breach` |
| `_check_keys` tolerates an absent `measured_t` | RED | `test_a_missing_measurement_time_is_a_rejected_record_not_a_silent_none` |
| stub never stamps `measured_t` | RED | 2 tests |
| stub accepts any pose age | RED | `test_an_unusable_pose_age_is_refused_by_the_transport` |
| pose age applied as `t + age` | RED | 2 tests, incl. the conformance run at `pose_age_seconds=2.0` |
| `_optional_float` maps `null` to `0.0` | RED | 2 round-trip tests |

Worth noting on row 3: `_RewindingTransport`
(`tests/test_transport_conformance.py:201`) is a *stalled* clock, "Reports
every tick at t=0.0", despite its name — so the `>` to `>=` mutation is
load-bearing on exactly the failure the rewritten justification names, not on
a rewind.

**The remaining rewritten prose.** Checked against the code rather than
accepted: `whiteout/transport/base.py:109-112` and
`tests/transport_conformance.py:254-255` on what `log.py` accepts — true,
`whiteout/log.py:280` is `record.t < previous`.
`tests/transport_conformance.py:366` on a non-finite `measured_t` being
unwritable — true, `_encode` passes `allow_nan=False` (`whiteout/log.py:112`)
and `_non_finite_field` (`whiteout/log.py:84`) rejects it on read.
`tests/transport_conformance.py:367-368` on `-inf` satisfying `<= t` — true,
and demonstrated by the planted breach. `tests/test_transport_conformance.py:28`
"Five of the eighteen cases" — `_PLANTED_BREACHES` now holds exactly 18
entries. Only the two Criticals above are false.

**Fixture regeneration.** Both claims are exact. Stripping `"measured_t":null,`
from each of the 120 new lines reproduces the `origin/main` file with zero
differing lines, and `run --seed 7 --ticks 120` here reproduces the committed
fixture byte for byte (119220 bytes both ways).

**As a contract for `arena`.** Per-pose `measured_t` is the right shape for
asynchronous MAVLink poses — each pose carries its own, so four assets
reporting at different instants land in one coherent observation without any
of them lying about the tick. `None` genuinely forces the question. The only
thing still pushing toward a wrong design is R1-M2, the unstated clock domain,
and the only thing the contract cannot express is a fix whose age is known but
whose absolute time is not — which nobody has asked for.

---

Ollienel777:
loop: fixer dispatched @ 145b830 — findings

---

Ollienel777:
fix: pushed @ e82ec7c

**R1-C1 — fixed.** The `SPEC.md` §4 citation is removed from the `Pose`
docstring, which now states the argument on its own terms: relaxing the
equality would leave the field, its type and its plausible value unchanged
while changing its meaning, so no test and no reader can tell the two apart.
`hackathon/SPEC.md` is not touched by this diff. Confirmed here that §4 is
"Architecture" (lines 153-216) and that `grep -n "cannot detect\|error class"`
over the file returns nothing, at this head and on `origin/main`. Whether §4
should gain that passage is now a follow-up in the PR body rather than
something this PR does.

**R1-C2 — fixed.** `check_the_tick_clock_is_monotonic`'s docstring no longer
quotes `base.py` as saying the clock is monotonic. It now names only
`whiteout/log.py` as the rule that admits equality, and says that `base.py`
states the same strict increase and that this check is what holds a transport
to it.

**R1-M1 — fixed, and the bump is taken.** `SCHEMA_VERSION` goes 1 → 2, with
`viz/viewer.js` (a test pins the two together) and
`fixtures/episodes/demo.jsonl` following. A log of the previous shape now
reads `line 1: schema_version 1 is not supported, expected 2` instead of
`record.observation.poses[0]: missing field(s) measured_t`, pinned by
`test_a_log_of_the_shape_before_measured_t_is_refused_by_version`. The
`SCHEMA_VERSION` docstring records why the round-trip rule is not the rule.
The PR body's attribution to #80 is removed — #80 says nothing about
`SCHEMA_VERSION`.

On #78: both branches change the same line to the same number, so they would
merge **without a conflict** and land `main` at 2 after two shape changes. The
PR body now opens with that, and says that if both land the final number must
be **3**, in `log.py` and `viewer.js`, with #78's fixture regeneration taken
over this one's. If only one lands, 2 stands.

**R1-M2 — accepted; the domain is stated.** `Pose` and `Transport.observe`
both now say `measured_t` is in our tick timebase, the same clock as `t`, and
that a far-side stamp is the field's *input*, never its value: an adapter
measures the offset between that clock and ours and subtracts it, or reports
`None`. `base.py`'s "read the far side's time only where it is the answer to a
different question" is followed immediately by that, so the sentence that sent
an author to the far side's clock no longer ends there. Nothing new is
asserted in the suite — a wrong-domain value that lands under `t` is not
detectable from in-process, which is why it is closed in the contract.

**R1-M3 — fixed.** `pose_age_seconds` is threaded through `create_transport`
and exposed as `run --pose-age SECONDS`. `run --seed 7 --ticks 3 --pose-age
2.0` writes records carrying `"measured_t":-2.0`, which `score` reads. The
default forwards nothing, so an episode without the flag is byte-for-byte what
it was — against the previous commit the regenerated fixture differs in
`schema_version` alone, and every pose in it is still `measured_t: null`. A
transport that does not take an age refuses it by name; the refusal is read
off the signature rather than off a caught `TypeError`, so a constructor's own
`TypeError` is not reported as a missing argument.

**R1-N1 — fixed.** `KinematicTransport` rejects a non-number before widening
it. Now: `'soon'`, `'2.0'` and `True` each raise `TransportError:
kinematic transport: pose_age_seconds ... is not a number of seconds`; `2.0`
is accepted.

**R1-N2 — not fixed here, reported.** `hackathon/SPEC.md:201` still lists
`Pose` without `measured_t`. It is a harness path, and editing it would make
this PR need a human merge, which it does not currently. It is in the PR
body's follow-ups, noting that #78 already edits `SPEC.md` and already needs a
human merge, so it should ride with that PR.

**Mutations — seven run, all RED.** `PYTHONDONTWRITEBYTECODE=1`, every
`__pycache__` cleared before and after each run, `-p no:cacheprovider`, and
the gate's own `whiteout source:` line confirming this worktree.

| mutation | result | caught by |
|---|---|---|
| kinematic drops the pre-coercion type guard | RED | `test_a_pose_age_that_is_not_a_number_is_refused_as_a_transport_error` |
| `SCHEMA_VERSION` back to 1 | RED | 2 tests, incl. the viewer-parity test |
| `create_transport` ignores the pose age | RED | 4 tests |
| `create_transport` defaults the pose age to `0.0` | RED | 2 tests |
| `run` stops forwarding `--pose-age` | RED | 2 tests |
| refusal read off a caught `TypeError`, not the signature | RED | `test_a_constructors_own_type_error_is_not_reported_as_a_missing_pose_age` |
| `viewer.js` version left at 1 | RED | `test_the_viewer_reads_the_same_schema_version_as_the_python_reader` |

Restored tree: `333 passed`.

**Gate, run here at e82ec7c, verbatim:**

```
=== [1/8] install ===
GATE PASS: install (0.1s)
=== [2/8] lint ===
GATE PASS: lint (0.0s)
=== [3/8] format ===
GATE PASS: format (0.0s)
=== [4/8] typecheck ===
GATE PASS: typecheck (0.2s)
=== [5/8] test ===
GATE PASS: test (3.4s)
=== [6/8] build ===
GATE PASS: build (0.5s)
=== [7/8] smoke ===
GATE PASS: smoke (0.5s)
=== [8/8] determinism ===
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.9s
```

---

Ollienel777:
loop: MERGE PRECONDITION — this PR may not merge while it reads
`SCHEMA_VERSION = 2`

Verified by the orchestrator, not taken from the fix reply:

    origin/main                          whiteout/log.py:63  SCHEMA_VERSION = 1
    origin/chore/73-one-coordinate-frame whiteout/log.py:69  SCHEMA_VERSION = 2   (PR #78)
    origin/chore/80-pose-timestamp-contract whiteout/log.py:76 SCHEMA_VERSION = 2   (this PR)

**Two independent record-shape changes both claim version 2**, on different
lines, so git will most likely merge them without a conflict and `main` would
end at 2 having changed shape twice. The version field exists precisely to say
"this file is not the shape you expect", and at 2 it would be unable to say it
about the other branch's logs — which is the same defect this PR's own R1-M1
was raised to fix, one level up.

**The reconciliation:**

- PR #78 is settled and awaiting a human merge. It goes first.
- **This PR must read `SCHEMA_VERSION = 3`** in `whiteout/log.py` *and* in
  `viz/viewer.js` (pinned by the parity test) before it merges, with
  `fixtures/episodes/demo.jsonl` regenerated on top of #78's version of that
  file rather than this branch's.
- The loop will not merge this PR until it does. This comment is the record of
  that condition; it survives a compaction, which a plan in a PR body does not.

The fix reply notes the reconciliation "rests on whoever merges second reading
the PR body, since git will not produce a conflict to force it". That is
correct, and it is why this is recorded as a precondition on the PR rather than
left in the body: **the loop merges this one, so the loop owns the check.**

---

Ollienel777:
semi-cold: closes @ e82ec7c

---

Ollienel777:
loop: #78 has merged — this PR is now `dirty` and the schema reconciliation is live

`main` is at `5d194ea` with `SCHEMA_VERSION = 2`. This PR also sets 2, and git
did **not** silently take it: `mergeable=false`, `mergeable_state=dirty`. The
conflict is real, which is better than the clean auto-merge that was feared.

**The precondition recorded at issuecomment-5745405274 now applies and can be
satisfied.** When a fixer next touches this branch it must:

1. merge `origin/main` in (never rebase) and resolve the conflict
2. set `SCHEMA_VERSION = 3` in `whiteout/log.py` **and** `viz/viewer.js`
   (pinned by the parity test) — 2 is now taken by #78's shape
3. regenerate `fixtures/episodes/demo.jsonl` on top of **main's** copy, not
   this branch's
4. update the message a v1 or v2 log fails with, so the version still says
   "this is not the shape you expect"

**No fixer is being dispatched yet**: a semi-cold round is in flight on this
PR at `e82ec7c`, and the brief forbids two subagents on one PR at once. The
conflict fix goes out when that round reports.

The round's verdict is still meaningful — it is judging the findings at
`e82ec7c`, and the conflict is with `main`, not with those findings.

---

Ollienel777 (commented):
semi-cold: closes @ e82ec7c

Round 2. Every verdict below is from running the code at e82ec7c in my own
worktree, `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`, every
`__pycache__` cleared between runs, and `whiteout.__file__` confirmed as
`/…/agent-a9f67514110e4666f/whiteout/__init__.py` (pinned on `PYTHONPATH`, not
installed). Gate here: `GATE OK: 8 steps in 7.1s`, all eight PASS. CI on
e82ec7c: `gate completed success`. Restored tree: `333 passed`.

The changed-file list against `origin/main` is 13 paths, **none of them a
harness path**: `fixtures/episodes/demo.jsonl`, `viz/viewer.js`,
`whiteout/{cli,log,types}.py`, `whiteout/transport/{__init__,base,kinematic}.py`
and five test files. No `docs/`, `.claude/`, `CLAUDE.md`, `.github/`,
`hackathon/`. The claim to an unattended merge holds.

---

## Verdicts

**R1-C1 — closed.** The clause is gone. `whiteout/types.py:279-287` now argues
from the mechanism itself ("No test and no reader can tell those two apart…")
with no citation attached, and **no replacement citation was invented**: the
added lines of the whole PR, filtered for `SPEC.md` or `§`, come back *empty* —
this PR adds no `SPEC.md` reference at all, in any file. The one `SPEC.md` §4
line still near it, `tests/transport_conformance.py:294` ("§4 makes
non-detection the input to the negative-information update"), predates the
branch — it is at `origin/main:tests/transport_conformance.py:283` verbatim and
is a different, true claim. `hackathon/SPEC.md` is untouched.

Worth recording, since the false claim in question originated upstream of the
fixer: **#80's own text carries it** ("precisely the error class `SPEC.md` §4
now records"), and of the two remedies round 1 offered, the other one —
adding the passage to `SPEC.md` — was not actually available: `SPEC.md` is a
harness path (`docs/build/RULES.md:79`) and `docs/build/TICKETS.md:62` is
"**Never edit `SPEC.md`.** If the spec is wrong, report it." Dropping the
citation and reporting it in the follow-ups was the only permitted move, and
it is the one taken.

**R1-C2 — closed.** Checked against what the two files now say, not against
the reply. `tests/transport_conformance.py:252-266` cites **only** `log.py`
for the equality-admitting rule ("`whiteout/log.py` accepts a repeated `t` and
refuses only one that runs backwards, so a stalled clock would pass there"),
then states that `base.py` "states that same strict increase on
`Transport.observe`, and this check is what holds a transport to it". Both
halves verified:

- `whiteout/log.py:293` is `if previous is not None and record.t < previous:`
  — a strict less-than, so a repeated `t` is accepted. The clause is true.
- `whiteout/transport/base.py:103-104` is "**`t` is our decision cadence, not
  the world's clock, and it strictly increases.**", and `base.py:71-72` says
  the clock `observe` returns "strictly increases for the life of the
  instance". A grep for `monotonic` in `base.py` has exactly one hit, line
  115, about the adapter's own counter — the same shape as before, and the
  docstring no longer claims otherwise.

The structural problem is gone too: the argument no longer says "neither of
the rules it sits next to is the reason" while `base.py` is one of them.

**R1-M1 — closed.** All three claims and the message verified by running the
reader:

```
SCHEMA_VERSION = 2
v1 line (from origin/main's fixture) -> line 1: schema_version 1 is not supported, expected 2
this branch's log read by a version-1 reader -> line 1: schema_version 2 is not supported, expected 1
```

The inverse direction — the one round 1 called worse — now reports the build
rather than `unknown field(s) measured_t`. `viz/viewer.js:35` is
`var SCHEMA_VERSION = 2;` and its own mismatch string (lines 120-124) prints
the same sentence the Python reader does. The fixture is regenerated: 120
lines, `schema_version` in {2}, all 480 `measured_t` values `None`, and
`run --seed 7 --ticks 120` here reproduces it **byte for byte**. Against
145b830 the fixture differs in `schema_version` alone.

The bump over an optional-on-read is the right call for the stated reason: an
optional read would collapse "absent" into "null", and absent-vs-null is the
whole content of the field.

**R1-M2 — closed, and the no-assertion reasoning holds.** The domain is now
stated in both places an adapter author lands: `whiteout/types.py:294-304` and
`whiteout/transport/base.py:131-139`, the latter placed immediately after the
sentence that used to end at "read the far side's time only where it is the
answer to a different question" — which was the actual wrong turn. Both say
the far-side stamp is the field's input, that the adapter subtracts the
measured offset, and that `None` is the answer when it cannot.

I went looking for the checkable thing you say does not exist, and I agree it
does not. The detectable cases are already detected — an epoch or a
milliseconds-not-seconds value is enormous and dies on `measured_t <= t` at
`tests/transport_conformance.py:390` on tick one. The undetectable case is a
recently-booted vehicle's small seconds value, and it is undetectable *in
principle* from in-process: a constant clock offset produces a constant
`t - measured_t`, which is indistinguishable from a constant genuine fix age
without a second time source we do not have. A staleness bound would be an
arbitrary number pretending to be a check. Closing it in the contract is the
honest outcome.

**R1-M3 — closed.** Re-run end to end here, not read:

```
$ python -m whiteout.cli run --seed 7 --ticks 3 --pose-age 2.0 --out …/age.jsonl
 t= 0.0 sv= 2 measured_t= [-2.0]
 t= 0.5 sv= 2 measured_t= [-1.5]
 t= 1.0 sv= 2 measured_t= [-1.0]
$ python -m whiteout.cli score …/age.jsonl --weights fixtures/weights/equal.json
total: 0.0000 (3 records)
```

`pose.t == record.t` holds on every line while `measured_t` sits 2.0 behind
it, the log validates, and the negative value at t=0 is the case
`types.py:295` says is legal. The default still forwards nothing (fixture: all
`measured_t: null`, and `--pose-age` absent reproduces the old bytes).

The signature-based refusal does what it claims. Probed directly against
patched factory tables:

```
factory without pose age, age given    : TransportError -> the 'kinematic' transport does not take a pose age
factory without pose age, no age       : OK -> NoAge
constructor raises its own TypeError   : TypeError -> my own unrelated TypeError
real kinematic, age 2.0                : OK
```

The third line is the point: the constructor's own `TypeError` propagates
intact and is *not* relabelled. A caught-`TypeError` implementation cannot
tell those apart, which is why `inspect.signature(...).bind` is the right
mechanism and not over-engineering.

**R1-N1 — closed.** `whiteout/transport/kinematic.py:89-95` guards before the
widening. Observed: `'soon'`, `'2.0'` and `True` each raise
`TransportError: kinematic transport: pose_age_seconds … is not a number of
seconds`; `-1.0` still raises the finiteness/sign error; `2.0` is accepted.
`'2.0'` is the one round 1 did not list and the one I would most expect from a
caller — it is refused.

**R1-N2 — withdrawn.** Not on the fixer's reasoning but on the rule: round 1
asserted `SPEC.md` "is not one of the harness files `CLAUDE.md` holds for a
human merge", and that was wrong. `docs/build/RULES.md:79` lists
`hackathon/SPEC.md` (once the spec PR has merged) among the paths that wait
for a human merge, and `docs/build/TICKETS.md:62` is flat: "Never edit
`SPEC.md`." So this was never a finding this PR could close, and deferring is
not merely defensible — it is the only compliant option. The residual work is
real and correctly parked in the follow-ups to ride with #78, which already
edits `SPEC.md` and already needs a human merge.

---

## Mutations — 7 run here, all RED

Independently applied and reverted, full suite each time.

| mutation | result | caught by |
|---|---|---|
| `SCHEMA_VERSION` back to 1 | RED | `test_the_viewer_reads_the_same_schema_version_as_the_python_reader`, `test_the_bundled_episode_is_a_real_valid_episode_log` |
| `viewer.js` left at 1 | RED | `test_the_viewer_reads_the_same_schema_version_as_the_python_reader` |
| `create_transport` ignores the pose age | RED | 5 tests, across `test_cli.py` and `test_transport.py` |
| `run` stops forwarding `--pose-age` | RED | `test_run_writes_a_stale_fix_into_the_log_when_asked`, `test_run_diagnoses_an_unusable_pose_age` |
| `--pose-age` defaults to `0.0` | RED | `test_run_without_a_pose_age_reports_no_measurement_time` |
| kinematic drops the pre-coercion type guard | RED | `test_a_pose_age_that_is_not_a_number_is_refused_as_a_transport_error` |
| refusal off a caught `TypeError`, not the signature | RED | `test_a_constructors_own_type_error_is_not_reported_as_a_missing_pose_age` |

No mutation came back green, and no failure landed on a neighbour of the
assertion it targets.

---

## On the #78 reconciliation

Not a finding, and I am not asking for 3 here — that would be wrong until #78
lands. The handling is **adequate, and better than the PR body alone
suggests**: the renumber is self-enforcing by the suite. I set
`SCHEMA_VERSION = 3` in `log.py` alone and the tree goes red on exactly the
two things that must move with it —
`test_the_viewer_reads_the_same_schema_version_as_the_python_reader` (the
viewer) and `test_the_bundled_episode_is_a_real_valid_episode_log` (the
fixture). A partial renumber at merge time cannot pass the gate.

`test_a_log_of_the_shape_before_measured_t_is_refused_by_version` also
survives the renumber intact: it builds its old record from
`SCHEMA_VERSION - 1` rather than a literal 1, so at 3 it still asserts the
mechanism instead of needing an edit.

---

## New in the commit range

Nothing above nit. The 12 new tests each pin a distinct behaviour, and the
three CLI ones are the ones that make #80's motive true rather than its
wording — `test_run_writes_a_stale_fix_into_the_log_when_asked` runs
`validate_episode_log` on the result and asserts `pose["t"] == record["t"]`
alongside the 2.0 offset, so it would catch a fix that bought staleness by
moving the tick.

### R2-N1 — `viz/viewer.js:320-322` still explains itself with "schema 1"

```
/* seed, transport, policy and terrain are episode provenance, and schema
 * 1 does not carry them — so the slots read as unknown rather than as a
 * value the viewer invented. */
```

Three lines above the version this PR just moved to 2, and the viewer now
*refuses* a schema 1 log outright (line 120). The behaviour is right — schema
2 does not carry provenance either; I checked the fixture's record keys:
`belief_digest, contacts, intent, observation, schema_version, t, truth` — but
the sentence now justifies the empty slots by a version this build will not
read. One number, or "the schema does not carry them". Never blocks, and it
will want touching again at 3.

### R2-N2 — a `**kwargs` factory would swallow a pose age instead of refusing it

`inspect.signature(factory).bind(seed=…, pose_age_seconds=…)` binds
successfully against `def __init__(self, **kw)`, so such a factory is handed
an age it may silently ignore, and `create_transport` reports success while
nothing is stamped. Demonstrated here with a `**kwargs` stand-in in the
factory table: `OK -> KwargsOnly`.

Remote — `arena` will almost certainly declare its arguments, and the realistic
case (an adapter that simply does not model fix age) is exactly what
`test_a_transport_that_does_not_take_a_pose_age_refuses_it_by_name` covers.
Recording it only because `arena` is written against this seam next and the
failure would be the quiet kind. Never blocks; not worth a change now.

---

Ollienel777:
loop: fixer dispatched @ e82ec7c — conflict

---

Ollienel777:
fix: pushed @ af7a743

`origin/main` merged in (never rebased). Three files conflicted:

- **`whiteout/types.py`** — `Pose.__post_init__`. Both sides rewrote the same line: main coerced `lat`/`lon` and added `_check_position`, this branch coerced `x`/`y` and added the optional `measured_t` coercion. Resolved to carry both: `_as_floats(self, "t", "lat", "lon", "z", "heading", "speed", "energy_used")`, the `measured_t is not None` coercion, then `_check_position(self, "pose")`. The field list and `from_dict` had already auto-merged with `lat`, `lon` and `measured_t` all present.
- **`whiteout/log.py`** — the `SCHEMA_VERSION` doc comment. Resolved by keeping both entries as separate versions rather than choosing one: 2 is the lat/lon frame of record from #78, 3 is `Pose.measured_t`. The "any change to the record shape" paragraph from this branch is kept above them.
- **`fixtures/episodes/demo.jsonl`** — every line. Reset to main's copy first, then regenerated with `run --seed 7 --ticks 120`.

Two tests this branch added still constructed a `Pose` with `x=`/`y=` and failed on the merged tree with `TypeError: Pose.__init__() got an unexpected keyword argument 'x'`. Carried onto `lat=72`/`lon=-95` in a second commit; the assertions are unchanged.

**Final `SCHEMA_VERSION` is 3**, in `whiteout/log.py:85` and `viz/viewer.js:35`.

**Fixture:** regenerated on top of main's copy and byte-identical to a fresh `run --seed 7 --ticks 120` (`cmp` clean). Every line reads `schema_version: 3`; pose keys are `asset_id, cls, energy_used, heading, lat, lon, measured_t, speed, t, z`, with `measured_t: null` throughout since the run passes no `--pose-age`.

**Failure messages.** Both readers check the version before any field check, and both name what they got and what they expected, so a v1 (x/y) *or* a v2 (lat/lon, no `measured_t`) log is refused by version and not by a nested missing field:

- `whiteout/log.py:257` — `schema_version 2 is not supported, expected 3`
- `whiteout/log.py:206` (writer) — `refusing to write schema_version 2, this build writes 3`
- `viz/viewer.js:123` — same text, same order, version checked before `RECORD_FIELDS`

`test_a_log_of_the_shape_before_measured_t_is_refused_by_version` is written against `SCHEMA_VERSION - 1` and is unedited; after the renumber its payload is a v2 log, and it still asserts `measured_t` does not appear in the message.

**Gate**, from a cleared `__pycache__` with `PYTHONDONTWRITEBYTECODE=1`:

```
  whiteout source: <not installed>
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (3.6s)
GATE PASS: build (0.5s)
GATE PASS: smoke (0.4s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 5.0s
```

No ambient `whiteout` distribution exists in this environment, so the only importable one is this worktree's; `whiteout.__file__` resolved here before the fixture was regenerated.

**Mutation, to show the renumber is guarded.** `log.py` back to 2, viewer left at 3:

```
FAILED tests/test_viz_shell.py::test_the_bundled_episode_is_a_real_valid_episode_log
FAILED tests/test_viz_shell.py::test_the_viewer_reads_the_same_schema_version_as_the_python_reader
2 failed, 353 passed
```

`viewer.js` back to 2, `log.py` at 3:

```
FAILED tests/test_viz_shell.py::test_the_viewer_reads_the_same_schema_version_as_the_python_reader
1 failed, 354 passed
```

Both reverted; the tree is clean at `af7a743`.

**Nits.** R2-N1 taken — `viz/viewer.js` `renderMeta` now says the record shape does not carry seed/transport/policy/terrain at any version the viewer reads, rather than naming schema 1. R2-N2 not taken: narrowing the `inspect.signature(...).bind` check to reject a `**kwargs` factory is a behaviour change plus a test, which is outside the merge and the renumber. It stands as recorded for `arena`.

**Diff scope.** The changed-file list against main is 13 files, all under `fixtures/`, `tests/`, `viz/` and `whiteout/`. No `docs/`, `.claude/`, `CLAUDE.md`, `.github/`, `hackathon/EVENT.md`, `hackathon/SPEC.md`, `scripts/setup-labels.sh` or `scripts/toutc.mjs`.

---

Ollienel777:
cold: clean @ af7a743

Round 3, settling. I read rounds 1 and 2 and the conflict fixer's reply first
and re-raise nothing that was closed. My attention went where it was pointed:
`git diff e82ec7c..af7a743` — the merge of `main` (#78, lat/lon, schema 2) and
the renumber to 3 — which no reviewer had seen. Everything below was run in my
own worktree at `af7a743`, `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`,
`__pycache__` cleared between runs, and `whiteout.__file__` confirmed as
`/…/agent-af4c8f60f328a1124/whiteout/__init__.py` (pinned on `PYTHONPATH`, and
the gate reports `whiteout source: <not installed>`).

**Gate here at `af7a743`, verbatim:**

```
=== [1/8] install ===
  whiteout source: <not installed>
GATE PASS: install (0.1s)
=== [2/8] lint ===
GATE PASS: lint (0.0s)
=== [3/8] format ===
GATE PASS: format (0.0s)
=== [4/8] typecheck ===
GATE PASS: typecheck (0.1s)
=== [5/8] test ===
GATE PASS: test (3.7s)
=== [6/8] build ===
GATE PASS: build (0.5s)
=== [7/8] smoke ===
GATE PASS: smoke (0.4s)
=== [8/8] determinism ===
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 5.1s
```

Restored tree: `355 passed`.

---

## The merge resolution: nothing was dropped, and I can show it

`bc44384` is a true merge (`Merge: e82ec7c 5d194ea`), never a rebase, and
`origin/main` is at `5d194ea` — fully merged in, nothing left to reconcile.

The decisive check is not reading the conflict hunks, it is reading the **net
delta against `main`**. `git diff origin/main...HEAD -- whiteout/types.py` is
**purely additive**: one field (`measured_t: float | None = None`), two lines
in `__post_init__`, one `_optional_float` call in `from_dict`, one helper, and
docstring. **Not one line of `main`'s is deleted or altered.** So every
position validation #78 added — `_check_position` on `Pose`,
`SensorFootprint`, `Detection`, `WaypointIntent`, `BeliefDigest`, `Contact`
and `TargetTruth`, and the `lat`/`lon` entries in each `_as_floats` — survives
by construction, not by the fixer's judgement. Same for `whiteout/log.py`: the
only deletion is this branch's own prose being reflowed around `main`'s.

The merged `whiteout/types.py:337-341` carries all three sides, in the right
order:

```python
_as_floats(self, "t", "lat", "lon", "z", "heading", "speed", "energy_used")
if self.measured_t is not None:
    _as_floats(self, "measured_t")
_check_position(self, "pose")
```

Order matters and is correct: `_as_floats` widens `lat`/`lon` before
`check_geodetic` reads them, so `Pose(lat=72, lon=-95)` with ints validates
rather than tripping `math.isfinite` on a non-float.

**Mutations — 5 run here, all RED.** Applied and reverted one at a time, full
suite each time.

| mutation | result | caught by |
|---|---|---|
| `_check_position(self, "pose")` deleted from `types.py:341` | RED | `test_a_swapped_lat_lon_at_the_strait_is_refused_at_construction` |
| the `measured_t` coercion deleted from `types.py:339-340` | RED | `test_an_int_measurement_time_is_coerced_like_every_other_float` |
| `"lat", "lon"` removed from `Pose`'s `_as_floats` | RED | `test_an_int_in_a_float_field_serialises_as_a_float` |
| reader's version check disabled (`log.py:254`) | RED | `test_validator_rejects_a_wrong_schema_version_naming_the_line`, `test_a_log_of_the_shape_before_measured_t_is_refused_by_version` |
| `SCHEMA_VERSION` back to 2, viewer left at 3 | RED | `test_the_viewer_reads_the_same_schema_version_as_the_python_reader`, `test_the_bundled_episode_is_a_real_valid_episode_log` |

No mutation came back green, and none failed on a neighbour of the assertion
it targets.

---

## `SCHEMA_VERSION - 1` still means what the test's name says

This is the one I expected to find something in, and it holds — for a
structural reason, not by luck.

`tests/test_types_log.py:301-304` builds `make_record(0).to_dict()`, labels it
`SCHEMA_VERSION - 1` = 2, and deletes `measured_t` from every pose.
**`measured_t` is the only field version 3 added over version 2**, so a v3
record with `measured_t` removed is byte-for-byte the v2 record shape. The
payload is therefore a genuine historical artifact, exactly as the test's name
claims: "the shape before `measured_t`" and "version 2" are the same shape
today.

It also still fails for the right reason, which is the part a passing test
cannot tell you. With the version check at `whiteout/log.py:254` disabled, it
goes red on precisely the message its docstring names:

```
AssertionError: assert 'schema_version' in
  'line 1: record.observation.poses[0]: missing field(s) measured_t'
```

So the assertion is load-bearing on the version mechanism, not on a
coincidence. Leaving it unedited was right.

---

## Both old shapes are refused by version, before any field check

Built from the real artifacts, not from hand-written payloads —
`5d194ea^:fixtures/episodes/demo.jsonl` is a true v1 log (`peak_xy`, x/y
poses) and `origin/main:fixtures/episodes/demo.jsonl` a true v2 one.

```
whiteout: /…/agent-af4c8f60f328a1124/whiteout/__init__.py   SCHEMA_VERSION: 3
v1.jsonl -> line 1: schema_version 1 is not supported, expected 3
v2.jsonl -> line 1: schema_version 2 is not supported, expected 3
```

Both name got and expected. `whiteout/log.py:252-258` runs the version check
before `_non_finite_field` and before `EpisodeRecord.from_dict`, so neither
log ever reaches a nested missing-field or unknown-field error. The writer
side is the same sentence at `whiteout/log.py:203-208`
(`refusing to write schema_version …, this build writes 3`). No message edit
was needed, as claimed.

I drove `viz/viewer.js` under `node` against all three logs rather than
trusting the pinning test:

```
viewer SCHEMA_VERSION = 3
viewer parsed the bundled episode: 120 records; pose keys:
  asset_id,cls,energy_used,heading,lat,lon,measured_t,speed,t,z
v1 -> line 1: schema_version 1 is not supported, expected 3
v2 -> line 1: schema_version 2 is not supported, expected 3
```

Same text, same order — `viz/viewer.js:121-126` checks the version before
`RECORD_FIELDS`. The viewer validates only top-level record fields, so the new
`measured_t` key needs no viewer change, and the demo beat (open the bundled
episode) is intact.

---

## The fixture

Regenerated here, not read:

```
$ python -m whiteout.cli run --seed 7 --ticks 120 --out …/demo.jsonl
$ cmp …/demo.jsonl fixtures/episodes/demo.jsonl
CMP CLEAN                       (121980 bytes)
```

120 lines, `schema_version` in {3}, pose keys exactly
`asset_id, cls, energy_used, heading, lat, lon, measured_t, speed, t, z`, and
all 480 `measured_t` values `None` — the default forwards nothing.
`fixtures/episodes/demo.jsonl` is the only committed episode artifact, so
nothing else needed regenerating.

The `--pose-age` path still works on the lat/lon build, end to end:

```
$ python -m whiteout.cli run --seed 7 --ticks 3 --pose-age 2.0 --out …/age.jsonl
 t= 0.0  sv= 3  measured_t= [-2.0]  pose.t==record.t: True  lat0= 71.99356
 t= 0.5  sv= 3  measured_t= [-1.5]  pose.t==record.t: True  lat0= 71.99356
 t= 1.0  sv= 3  measured_t= [-1.0]  pose.t==record.t: True  lat0= 71.99356
$ python -m whiteout.cli score …/age.jsonl --weights fixtures/weights/equal.json
total: 0.0000 (3 records)
```

---

## The fixer's other claims, checked

- **The two ported tests.** `git show af7a743` is 4 lines across 2 tests:
  `x=0.0,y=0.0` to `lat=72.0,lon=-95.0`, and `x=0,y=0` to `lat=72,lon=-95`.
  **Every assertion is untouched**, including the int-coercion one, which is
  the one that would have been tempting to soften. `lat=72, lon=-95` is inside
  `check_geodetic`'s ranges, and the int arms still exercise the widening.
- **Residue I went looking for and did not find.** A grep over `whiteout/`,
  `tests/` and `viz/` for `target_xy`, `peak_xy`, `pose.x` and `x=`/`y=`
  constructions returns only `whiteout/log.py:74` and `whiteout/types.py:22`
  (both prose *about* the old frame, correctly in the past tense) and
  `tests/test_geo.py:66`'s ban list. No test still builds a record on the old
  frame.
- **The count in `tests/test_transport_conformance.py:28`** — "Five of the
  eighteen cases" — is still true after the merge: `_PLANTED_BREACHES` holds
  18 entries and `CONFORMANCE_CHECKS` 11. #78 added neither.
- **`enu_to_geodetic(pose.lat, pose.lon, 100.0, 100.0)`** at
  `tests/transport_conformance.py:432` matches the signature at
  `whiteout/geo.py:268` — `(lat_deg, lon_deg, east_m, north_m)`, returning a
  `GeoPoint`. Argument order is right; a swap would have been the easy defect
  here, and `test_every_check_owns_a_planted_breach` would not have caught it.

## Prose sweep

Every checkable claim the merge wrote or moved, verified against the code:

- `whiteout/log.py:74-78` — "``SPEC.md`` §5 names [lat/lon] as the frame of
  record". `hackathon/SPEC.md` §5 "Conventions" does: "**The frame of record
  is geodetic WGS-84 `lat`/`lon`, in degrees.**" **True**, and this PR does
  not touch `SPEC.md`.
- `whiteout/log.py:80-84` — "``Pose.measured_t``, a required record key".
  **True**: `_check_keys` requires every declared field, demonstrated by the
  disabled-version-check run above producing `missing field(s) measured_t`.
- `whiteout/log.py:81-84` — "It is 3 and not 2 because 2 is already spent on
  the lat/lon shape". **True**, and the two mutation rows above show the
  renumber is self-enforcing across `log.py`, `viewer.js` and the fixture.
- `whiteout/types.py:20-27` (kept from `main`) — "Construction range-checks
  every pair". **True** for all seven record types; verified additively and by
  mutation.
- `viz/viewer.js:322-324` — R2-N1 taken, and the replacement is true: no
  version of the record shape has ever carried seed/transport/policy/terrain,
  and the fixture's record keys confirm it.
- `whiteout/transport/base.py` and `tests/transport_conformance.py` prose is
  unchanged by the merge; I spot-checked that #78 did not falsify it — the
  package still imports nothing from `belief`/`policy`/`score`/`sim`, and
  `whiteout.geo`, which `kinematic.py` now imports, is not among the modules
  that sentence excludes.

**No false claim found.** The run's overclaim streak is not extended here.

## Diff scope

`git diff --name-only origin/main...HEAD` is 13 paths —
`fixtures/episodes/demo.jsonl`, `viz/viewer.js`, `whiteout/{cli,log,types}.py`,
`whiteout/transport/{__init__,base,kinematic}.py` and five test files.
**No `docs/`, `.claude/`, `CLAUDE.md`, `.github/`, `hackathon/EVENT.md`,
`hackathon/SPEC.md`, `scripts/setup-labels.sh` or `scripts/toutc.mjs`.**
Nothing crept in during the merge; the claim to an unattended merge holds.

---

## Nits — neither blocks

### R3-N1 — `tests/test_types_log.py:302` is right today and decays at the next bump

`payload["schema_version"] = SCHEMA_VERSION - 1` is exactly "the shape before
`measured_t`" **only because `measured_t` is the newest field**. The moment a
version 4 adds a field, `make_record(0)` carries it, `SCHEMA_VERSION - 1` is 3,
and the payload becomes "v4 minus `measured_t`" — a shape no build ever wrote,
labelled 3. The test keeps passing, and its name and docstring ("a log written
before it existed") quietly stop being true. It survived *this* renumber
intact, which is the good half of the design; it is the next one that would
turn it into a test that passes for a changed reason. A literal `2` with a
one-line comment would pin it, at the cost of the property round 2 praised.
Recording it rather than asking for it.

### R3-N2 — `whiteout/log.py:61-85` is now a version log with no entry for 1

The comment became a numbered list at this merge (2, then 3), but there is no
`1 —` entry, while line 76 still says "A version-1 line is refused rather than
read as version 2". Someone holding a v1 artifact — and the repo contains one
in its history — has nothing here telling them what 1 was. Inherited from
`main`'s shape and sharpened by the list; one line would close it. Never
blocks.

*(R2-N1 taken by the fixer. R2-N2 stands as recorded, deliberately out of
scope for a merge fix.)*

---

## Verdict

No new Critical and no new Medium. The merge resolution is correct, it drops
nothing from either side, the renumber is guarded by the suite in three
places, both prior shapes are refused by version with the right message before
any field check, the fixture reproduces byte for byte, and the diff touches no
harness path. **Settles.**

---

Ollienel777 (commented):
cold: clean @ af7a743

Round 3, settling. I read rounds 1 and 2 and the conflict fixer's reply first
and re-raise nothing that was closed. My attention went where it was pointed:
`git diff e82ec7c..af7a743` — the merge of `main` (#78, lat/lon, schema 2) and
the renumber to 3 — which no reviewer had seen. Everything below was run in my
own worktree at `af7a743`, `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`,
`__pycache__` cleared between runs, and `whiteout.__file__` confirmed as
`/…/agent-af4c8f60f328a1124/whiteout/__init__.py` (pinned on `PYTHONPATH`, and
the gate reports `whiteout source: <not installed>`).

**Gate here at `af7a743`, verbatim:**

```
=== [1/8] install ===
  whiteout source: <not installed>
GATE PASS: install (0.1s)
=== [2/8] lint ===
GATE PASS: lint (0.0s)
=== [3/8] format ===
GATE PASS: format (0.0s)
=== [4/8] typecheck ===
GATE PASS: typecheck (0.1s)
=== [5/8] test ===
GATE PASS: test (3.7s)
=== [6/8] build ===
GATE PASS: build (0.5s)
=== [7/8] smoke ===
GATE PASS: smoke (0.4s)
=== [8/8] determinism ===
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 5.1s
```

Restored tree: `355 passed`.

---

## The merge resolution: nothing was dropped, and I can show it

`bc44384` is a true merge (`Merge: e82ec7c 5d194ea`), never a rebase, and
`origin/main` is at `5d194ea` — fully merged in, nothing left to reconcile.

The decisive check is not reading the conflict hunks, it is reading the **net
delta against `main`**. `git diff origin/main...HEAD -- whiteout/types.py` is
**purely additive**: one field (`measured_t: float | None = None`), two lines
in `__post_init__`, one `_optional_float` call in `from_dict`, one helper, and
docstring. **Not one line of `main`'s is deleted or altered.** So every
position validation #78 added — `_check_position` on `Pose`,
`SensorFootprint`, `Detection`, `WaypointIntent`, `BeliefDigest`, `Contact`
and `TargetTruth`, and the `lat`/`lon` entries in each `_as_floats` — survives
by construction, not by the fixer's judgement. Same for `whiteout/log.py`: the
only deletion is this branch's own prose being reflowed around `main`'s.

The merged `whiteout/types.py:337-341` carries all three sides, in the right
order:

```python
_as_floats(self, "t", "lat", "lon", "z", "heading", "speed", "energy_used")
if self.measured_t is not None:
    _as_floats(self, "measured_t")
_check_position(self, "pose")
```

Order matters and is correct: `_as_floats` widens `lat`/`lon` before
`check_geodetic` reads them, so `Pose(lat=72, lon=-95)` with ints validates
rather than tripping `math.isfinite` on a non-float.

**Mutations — 5 run here, all RED.** Applied and reverted one at a time, full
suite each time.

| mutation | result | caught by |
|---|---|---|
| `_check_position(self, "pose")` deleted from `types.py:341` | RED | `test_a_swapped_lat_lon_at_the_strait_is_refused_at_construction` |
| the `measured_t` coercion deleted from `types.py:339-340` | RED | `test_an_int_measurement_time_is_coerced_like_every_other_float` |
| `"lat", "lon"` removed from `Pose`'s `_as_floats` | RED | `test_an_int_in_a_float_field_serialises_as_a_float` |
| reader's version check disabled (`log.py:254`) | RED | `test_validator_rejects_a_wrong_schema_version_naming_the_line`, `test_a_log_of_the_shape_before_measured_t_is_refused_by_version` |
| `SCHEMA_VERSION` back to 2, viewer left at 3 | RED | `test_the_viewer_reads_the_same_schema_version_as_the_python_reader`, `test_the_bundled_episode_is_a_real_valid_episode_log` |

No mutation came back green, and none failed on a neighbour of the assertion
it targets.

---

## `SCHEMA_VERSION - 1` still means what the test's name says

This is the one I expected to find something in, and it holds — for a
structural reason, not by luck.

`tests/test_types_log.py:301-304` builds `make_record(0).to_dict()`, labels it
`SCHEMA_VERSION - 1` = 2, and deletes `measured_t` from every pose.
**`measured_t` is the only field version 3 added over version 2**, so a v3
record with `measured_t` removed is byte-for-byte the v2 record shape. The
payload is therefore a genuine historical artifact, exactly as the test's name
claims: "the shape before `measured_t`" and "version 2" are the same shape
today.

It also still fails for the right reason, which is the part a passing test
cannot tell you. With the version check at `whiteout/log.py:254` disabled, it
goes red on precisely the message its docstring names:

```
AssertionError: assert 'schema_version' in
  'line 1: record.observation.poses[0]: missing field(s) measured_t'
```

So the assertion is load-bearing on the version mechanism, not on a
coincidence. Leaving it unedited was right.

---

## Both old shapes are refused by version, before any field check

Built from the real artifacts, not from hand-written payloads —
`5d194ea^:fixtures/episodes/demo.jsonl` is a true v1 log (`peak_xy`, x/y
poses) and `origin/main:fixtures/episodes/demo.jsonl` a true v2 one.

```
whiteout: /…/agent-af4c8f60f328a1124/whiteout/__init__.py   SCHEMA_VERSION: 3
v1.jsonl -> line 1: schema_version 1 is not supported, expected 3
v2.jsonl -> line 1: schema_version 2 is not supported, expected 3
```

Both name got and expected. `whiteout/log.py:252-258` runs the version check
before `_non_finite_field` and before `EpisodeRecord.from_dict`, so neither
log ever reaches a nested missing-field or unknown-field error. The writer
side is the same sentence at `whiteout/log.py:203-208`
(`refusing to write schema_version …, this build writes 3`). No message edit
was needed, as claimed.

I drove `viz/viewer.js` under `node` against all three logs rather than
trusting the pinning test:

```
viewer SCHEMA_VERSION = 3
viewer parsed the bundled episode: 120 records; pose keys:
  asset_id,cls,energy_used,heading,lat,lon,measured_t,speed,t,z
v1 -> line 1: schema_version 1 is not supported, expected 3
v2 -> line 1: schema_version 2 is not supported, expected 3
```

Same text, same order — `viz/viewer.js:121-126` checks the version before
`RECORD_FIELDS`. The viewer validates only top-level record fields, so the new
`measured_t` key needs no viewer change, and the demo beat (open the bundled
episode) is intact.

---

## The fixture

Regenerated here, not read:

```
$ python -m whiteout.cli run --seed 7 --ticks 120 --out …/demo.jsonl
$ cmp …/demo.jsonl fixtures/episodes/demo.jsonl
CMP CLEAN                       (121980 bytes)
```

120 lines, `schema_version` in {3}, pose keys exactly
`asset_id, cls, energy_used, heading, lat, lon, measured_t, speed, t, z`, and
all 480 `measured_t` values `None` — the default forwards nothing.
`fixtures/episodes/demo.jsonl` is the only committed episode artifact, so
nothing else needed regenerating.

The `--pose-age` path still works on the lat/lon build, end to end:

```
$ python -m whiteout.cli run --seed 7 --ticks 3 --pose-age 2.0 --out …/age.jsonl
 t= 0.0  sv= 3  measured_t= [-2.0]  pose.t==record.t: True  lat0= 71.99356
 t= 0.5  sv= 3  measured_t= [-1.5]  pose.t==record.t: True  lat0= 71.99356
 t= 1.0  sv= 3  measured_t= [-1.0]  pose.t==record.t: True  lat0= 71.99356
$ python -m whiteout.cli score …/age.jsonl --weights fixtures/weights/equal.json
total: 0.0000 (3 records)
```

---

## The fixer's other claims, checked

- **The two ported tests.** `git show af7a743` is 4 lines across 2 tests:
  `x=0.0,y=0.0` to `lat=72.0,lon=-95.0`, and `x=0,y=0` to `lat=72,lon=-95`.
  **Every assertion is untouched**, including the int-coercion one, which is
  the one that would have been tempting to soften. `lat=72, lon=-95` is inside
  `check_geodetic`'s ranges, and the int arms still exercise the widening.
- **Residue I went looking for and did not find.** A grep over `whiteout/`,
  `tests/` and `viz/` for `target_xy`, `peak_xy`, `pose.x` and `x=`/`y=`
  constructions returns only `whiteout/log.py:74` and `whiteout/types.py:22`
  (both prose *about* the old frame, correctly in the past tense) and
  `tests/test_geo.py:66`'s ban list. No test still builds a record on the old
  frame.
- **The count in `tests/test_transport_conformance.py:28`** — "Five of the
  eighteen cases" — is still true after the merge: `_PLANTED_BREACHES` holds
  18 entries and `CONFORMANCE_CHECKS` 11. #78 added neither.
- **`enu_to_geodetic(pose.lat, pose.lon, 100.0, 100.0)`** at
  `tests/transport_conformance.py:432` matches the signature at
  `whiteout/geo.py:268` — `(lat_deg, lon_deg, east_m, north_m)`, returning a
  `GeoPoint`. Argument order is right; a swap would have been the easy defect
  here, and `test_every_check_owns_a_planted_breach` would not have caught it.

## Prose sweep

Every checkable claim the merge wrote or moved, verified against the code:

- `whiteout/log.py:74-78` — "``SPEC.md`` §5 names [lat/lon] as the frame of
  record". `hackathon/SPEC.md` §5 "Conventions" does: "**The frame of record
  is geodetic WGS-84 `lat`/`lon`, in degrees.**" **True**, and this PR does
  not touch `SPEC.md`.
- `whiteout/log.py:80-84` — "``Pose.measured_t``, a required record key".
  **True**: `_check_keys` requires every declared field, demonstrated by the
  disabled-version-check run above producing `missing field(s) measured_t`.
- `whiteout/log.py:81-84` — "It is 3 and not 2 because 2 is already spent on
  the lat/lon shape". **True**, and the two mutation rows above show the
  renumber is self-enforcing across `log.py`, `viewer.js` and the fixture.
- `whiteout/types.py:20-27` (kept from `main`) — "Construction range-checks
  every pair". **True** for all seven record types; verified additively and by
  mutation.
- `viz/viewer.js:322-324` — R2-N1 taken, and the replacement is true: no
  version of the record shape has ever carried seed/transport/policy/terrain,
  and the fixture's record keys confirm it.
- `whiteout/transport/base.py` and `tests/transport_conformance.py` prose is
  unchanged by the merge; I spot-checked that #78 did not falsify it — the
  package still imports nothing from `belief`/`policy`/`score`/`sim`, and
  `whiteout.geo`, which `kinematic.py` now imports, is not among the modules
  that sentence excludes.

**No false claim found.** The run's overclaim streak is not extended here.

## Diff scope

`git diff --name-only origin/main...HEAD` is 13 paths —
`fixtures/episodes/demo.jsonl`, `viz/viewer.js`, `whiteout/{cli,log,types}.py`,
`whiteout/transport/{__init__,base,kinematic}.py` and five test files.
**No `docs/`, `.claude/`, `CLAUDE.md`, `.github/`, `hackathon/EVENT.md`,
`hackathon/SPEC.md`, `scripts/setup-labels.sh` or `scripts/toutc.mjs`.**
Nothing crept in during the merge; the claim to an unattended merge holds.

---

## Nits — neither blocks

### R3-N1 — `tests/test_types_log.py:302` is right today and decays at the next bump

`payload["schema_version"] = SCHEMA_VERSION - 1` is exactly "the shape before
`measured_t`" **only because `measured_t` is the newest field**. The moment a
version 4 adds a field, `make_record(0)` carries it, `SCHEMA_VERSION - 1` is 3,
and the payload becomes "v4 minus `measured_t`" — a shape no build ever wrote,
labelled 3. The test keeps passing, and its name and docstring ("a log written
before it existed") quietly stop being true. It survived *this* renumber
intact, which is the good half of the design; it is the next one that would
turn it into a test that passes for a changed reason. A literal `2` with a
one-line comment would pin it, at the cost of the property round 2 praised.
Recording it rather than asking for it.

### R3-N2 — `whiteout/log.py:61-85` is now a version log with no entry for 1

The comment became a numbered list at this merge (2, then 3), but there is no
`1 —` entry, while line 76 still says "A version-1 line is refused rather than
read as version 2". Someone holding a v1 artifact — and the repo contains one
in its history — has nothing here telling them what 1 was. Inherited from
`main`'s shape and sharpened by the list; one line would close it. Never
blocks.

*(R2-N1 taken by the fixer. R2-N2 stands as recorded, deliberately out of
scope for a merge fix.)*

---

## Verdict

No new Critical and no new Medium. The merge resolution is correct, it drops
nothing from either side, the renumber is guarded by the suite in three
places, both prior shapes are refused by version with the right message before
any field check, the fixture reproduces byte for byte, and the diff touches no
harness path. **Settles.**

---

Ollienel777:
settled: @ af7a743

---

PR #79, opened 2026-09-19
feat(belief): the ribbon belief field over the strait, and its detection update

Ollienel777:
Closes #12

Opened by the build loop.

## What

Adds `whiteout/belief/`: a normalised belief field over the water of Bellot
Strait, a diffusion step, a Bayesian detection update, and the protocol the
policy reads it through.

- `belief/geometry.py` — the strait as a **centreline polyline in lat/lon plus
  a half-width**. A point is on water when `abs(w) <= half_width(s)`.
  `DEFAULT_STRAIT` is four vertices tapering from 1000 m at each mouth to
  520 m at the narrows.
- `belief/grid.py` — `ChannelBeliefGrid`, the field itself: uniform over water
  to start, flux-form diffusion, and `update_detection`.
- `belief/field.py` — the `BeliefField` protocol. Positions and numbers only;
  no cells, no indices, no axes.

Nothing outside `whiteout/belief/` changes. No existing file is touched, and
the CLI, the log schema and `types.BeliefDigest` are all untouched.

## The two judgement calls the ticket delegates

**1. The shape of the field: a ribbon, not a raster.** Indexed by arc length
`s` along the centreline and signed offset `w` across it. On the default
geometry at 100 m resolution:

| grid | cells | water | useful |
|---|---|---|---|
| North/East raster over the ribbon's bounding box | 9072 | 3992 | 44 % |
| this ribbon, 251 along × 20 across | 5020 | 3992 | **79 %** |

Same 39.9 km² of water, 45 % fewer cells, and the gap widens with every degree
the strait bends. Cell count is the smaller half of the argument. The other
half:

- **The shoreline becomes an inequality**, not a terrain model. `ARENA.md` §7
  killed the terrain generator, so there is no heightfield to consult, and a
  half-width per vertex is the whole of what is needed.
- **The axes are the search's axes.** "Which stretch is unsearched?" is an
  interval in `s`; "belief is drifting east" is a statement about `s` alone.
  On a North/East raster both are diagonal queries. #13's negative-information
  update and #14's policy both want the along-channel marginal, and here that
  is one sum.

**2. The diffusion rate: σ = 43.8 m per axis per 1 s tick.** Set from the
shape of the loss, then derived:

- **Speed is a parameter, and its default is 8.0 m/s (about 15.5 kn).** There
  is no measurement of the vessel's speed to set it from. `ARENA.md` §5's
  `speed: 6.5` is **a field inside a payload we POST** — the example body of a
[...]

PR #83, opened 2026-09-19
chore(infra): let MAVLink's degE7 factor past the ellipsoid guard

### LINKED ISSUE #82: chore: the ellipsoid guard rejects MAVLink's degE7 factor
Filed by the build loop.

**Milestone:** M1 · **Demo beat:** 6 · **Size:** S

## What
`tests/test_geo.py`'s ellipsoid guard has **no allowlist**, and its range
`_ELLIPSOID_SCALE_M = (1e5, 1e8)` **contains `1e7`** — MAVLink's degE7 scaling
factor. So any literal `1e7` in `whiteout/` fails the gate with a message about
earth radii.

Demonstrated by PR #78's settling round with a probe file: `1e7` in
`whiteout/transport/` fails the gate, citing earth radii.

Fix so that the guard still catches a second ellipsoid constant while allowing
the degE7 conversion an adapter must perform. Options, in rough order of
preference:

1. Narrow the range so it excludes `1e7` while still bracketing the WGS-84
   semi-major axis and its relatives
2. Allow `1e7` specifically, by name, with a comment saying what it is
3. An allowlist entry per file, which is what the trigonometry guard already
   does — weakest, because it exempts the whole file

## Why
`GLOBAL_POSITION_INT` reports `lat` and `lon` as integers scaled by 1e7. **Every
MAVLink adapter must divide by it.** So the guard as merged will reject the
`arena` and `sitl` transports on sight — that is **#64 and #68**, and
`SPEC.md` §3 and §9 make the MAVLink transport **demo beat 6**.

It is harmless today only because nothing in the repo writes `1e7`. It becomes
a blocked ticket the moment something does, and the failure message points at
earth radii rather than at the guard's range, so whoever hits it will lose time
before finding the cause.

**Fix this before #64 or #68 is dispatched, not after.**

## Acceptance
- [ ] A file in `whiteout/transport/` containing MAVLink's `1e7` scaling passes
      the gate
- [ ] The guard still fails on a second copy of the ellipsoid constants — add a
      planted-breach test so this cannot silently reopen
- [ ] The guard's failure message names the guard and its range, so a future
      hit is diagnosable without reading `test_geo.py`
- [ ] `whiteout/geo.py` remains the only coordinate converter

## Notes
`tests/test_geo.py:147`, merged in PR #78 at `5d194ea`. Recorded there as
R3-M4 by the settling round, and not fixed under the review bar on #50 because
it breaks nothing at that head.
Related: #64 (tracks API client), #68 (tower control), #23 (arena adapter).

---

Ollienel777:
Closes #82

Opened by the build loop.

## What

`tests/test_geo.py`'s ellipsoid guard now lets two values through its
`(1e5, 1e8)` metre range, held in `_WIRE_FACTORS` with what each one decodes:
`1e7` (MAVLink degE7 lat/lon) and `1e6` (`time_usec` microseconds). Both are
dimensionless wire factors, and the concern is one sentence long: *the
metre-scale range must not adjudicate MAVLink wire factors.* Four further
changes around it:

- Each of the three static guards moves from a test body into a
  `_check_*(where, text)` function that the walking test drives. The walk is
  unchanged; the guards can now be run against source that is not on disk.
- Every guard failure names the guard (module path plus test name), the rule
  it applies — for the ellipsoid guard, its range — and the full exemption
  list. The old message said only "an earth radius or a metres-per-degree
  factor".
- 17 planted breaches, in the idiom of
  `tests/test_transport_conformance.py::_PLANTED_BREACHES`, plus
  `test_every_guard_owns_a_planted_breach` and
  `test_every_walking_guard_is_registered`, which together fail whether a
  guard is dropped from `_GUARDS` or never added to it.
- `_NOT_OURS` gains `.claude`, so `_sources()` stops descending into
  `.claude/worktrees/` (PR #78's R2-M2). The filter is lifted into `_is_ours`,
  a pure predicate, and pinned.

No file outside `tests/` changes. `whiteout/geo.py` is untouched.

**Which of the ticket's three options, and why.** Option 2 — allow the wire
factors by value, by name, with a comment — rather than option 1, narrowing
the range. Narrowing the top below `1e7` would also free everything from `1e7`
to `1e8`, and that decade is not empty: the equatorial circumference is
40 075 017 m and the polar meridian 40 007 863 m, and either divided by 360 is
a metres-per-degree factor — a second converter reached without ever writing
`111320`. Two planted breaches pin exactly that (`EARTH_CIRCUMFERENCE_M`,
`MERIDIAN_M`); both go green under option 1, which is mutation M3 below.
Option 3 was not considered further: it exempts a whole file, and PR #78's own
review appended a converter to an allowlisted file and passed all three
guards.

**Why `1e6` is here rather than in a follow-up.** `1e6 / 9` and `1e7 / 90` are
the same float (111 111.111), so every derived metres-per-degree a `1e6`
exemption can reach, the `1e7` exemption already reached — the marginal hole
is zero. And nothing geodetic lives near 1e6: `1e6 / 90` is 11 111 and
`1e6 / 360` is 2 778; no radius, no circumference, no quadrant — unlike 1e7,
which sits 2 km from the meridian quadrant and is exempted anyway. It is a
second entry in a tuple this PR creates, not a second concern, and leaving it
would make #64 and #68 open and defend a change to `tests/test_geo.py`
mid-event.

**What the exemption costs — measured at the strait, against
`whiteout/geo.py`.** The figures in the first revision of this PR were wrong
in both halves, and are corrected here (R1-C1):

- **The magnitude.** `1e7 / 90` = 111 111.1 is only 0.02% off the *global
  mean* meridian degree, but this guard exists to stop a converter disagreeing
  with `whiteout/geo.py`. At 71.99 N `radii_of_curvature` gives
  M = 6 393 414.1 m, i.e. 111 586.1 m/deg, so `1e7 / 90` is **0.426% low:
  about 4.3 m per km North, and 53.44 m at the arena's 12.5 km half-length**
  — 22x the "0.2 m per km" first claimed.
- **The containment.** The claim that "the longitude half still needs a
  cosine, which the trigonometry guard refuses" was **false**. A precomputed
  `COS_ORIGIN = 0.3093051` is neither an `ast.Attribute` nor a
  `from math import`, and is far below the range, so the whole two-axis
  converter — `lat + north / (1e7 / 90)`,
  `lon + east / ((1e7 / 90) * COS_ORIGIN)` — **passes all three guards**, with
  a worst disagreement of **53.63 m** over the arena. That is within 3 m of
  the 51 m converter option 1 was rejected to prevent.

Stated plainly, because a sub-metre number here is how the next widening gets
justified: **this exemption opens a two-axis converter that is ~54 m out at
the arena corner, and none of the three guards catches it.** Option 2 is still
the right call — it opens one door where option 1 opened a decade of them —
but the cost is ~54 m, not sub-metre. It is the same shape as PR #78's R2-N1
precomputed-cosine bypass, which is still open; closing that closes this too.

While correcting those, the neighbouring claim that the naive `111320.0`
"disagrees with `whiteout.geo` by about 90 m over the strait" turned out to be
wrong as well — measured against `geodetic_to_local`, the same two-axis
converter spelt with `111320.0` is **30.0 m** out at the worst corner. That is
now the stated figure, and it makes the honest point that the guard refuses
the less wrong spelling (30.0 m, caught) and admits the more wrong one
(53.63 m, exempt).

What the exemption does contain: it is two values, not two bands. Both
neighbours of both factors are planted (`10000001.0`, `9999999.0`,
`1000001.0`, `999999.0`), and the quadrant's own value (10 001 966 m) is
still in range.

## Why

`GLOBAL_POSITION_INT` reports `lat`/`lon` as integers of 1e-7 degrees and
`time_usec` in microseconds, so every MAVLink adapter divides by `1e7` and
`1e6`. As merged, the guard rejected the `arena` and `sitl` transports on
sight — #64 and #68, `SPEC.md` §3 and §9, demo beat 6.

## How to verify

```bash
git fetch origin chore/82-dege7-guard && git checkout chore/82-dege7-guard
python scripts/gate.py
python -m pytest tests/test_geo.py -q            # 41 passed
```

Per acceptance criterion:

| criterion | test |
|---|---|
| a file in `whiteout/transport/` containing `1e7` passes | `test_a_mavlink_adapter_may_spell_the_wire_factors` — runs all three guards over an adapter snippet at `whiteout/transport/mavlink.py`, a path in no allowlist, spelling both `1e7` and `1e6` |
| the guard still fails a second copy of the ellipsoid; planted breach | `test_a_guard_fails_the_source_it_exists_to_catch` (17 cases), `test_every_guard_owns_a_planted_breach` and `test_every_walking_guard_is_registered` |
| the failure message names the guard and its range | `test_a_guard_failure_names_the_guard_and_what_it_refuses`, `test_the_name_guard_says_it_matches_regardless_of_case` |
| `whiteout/geo.py` remains the only converter | the three walking tests, unchanged in what they walk |
| the `.claude` exclusion is what it claims | `test_the_exclusion_of_other_agents_worktrees_holds` — `_is_ours` is a pure predicate, so it pins with no filesystem |

The message a future hit gets, verbatim:

```
whiteout/transport/probe.py line 1 spells out 111320.0. The ellipsoid guard
(tests/test_geo.py::test_only_whiteout_geo_defines_the_ellipsoid) fails any
literal whose magnitude falls in [1e+05, 1e+08], because a constant that size
is an earth radius, a circumference or a metres-per-degree factor -- and the
ellipsoid lives in whiteout/geo.py and nowhere else. Its only exemptions are
1e+07 (degE7 lat/lon), 1e+06 (time_usec microseconds). If yours is another
dimensionless wire factor rather than a second frame, name it beside those in
test_geo.py's _WIRE_FACTORS; do not widen the range
```

**Mutations run, each confirmed red on the named test** (with
`PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`). M1–M8 were run against the
first revision; N1–N8 against this one.

| # | mutation | went red |
|---|---|---|
| M1 | drop the degE7 exemption entirely | `test_a_mavlink_adapter_may_spell_the_wire_factors` |
| M2 | widen the exemption from one value to a ±1e6 band | breach *upper neighbour of the degE7 exemption* |
| M3 | take option 1 instead — range top to `9.99e6` | breach *the equatorial circumference*, and *the polar meridian* |
| M4 | drop `_check_the_trigonometry` from `_GUARDS` | now `test_nothing_but_whiteout_geo_does_trigonometry_on_a_latitude`, `test_every_walking_guard_is_registered` **and** `test_every_guard_owns_a_planted_breach` — see N3 |
| M5 | drop both trigonometry planted breaches | `test_every_guard_owns_a_planted_breach` |
| M6 | collect only `ast.Assign` in `_defined_names` | breach *a conversion named by an annotated assignment* |
| M7 | failure message stops naming the guard | `test_a_guard_failure_names_the_guard_and_what_it_refuses` |
| M8 | failure message stops naming the range | `test_a_guard_failure_names_the_guard_and_what_it_refuses` |
| N1 | drop the `1e6` wire factor | `test_a_mavlink_adapter_may_spell_the_wire_factors` |
| N2 | widen the wire-factor exemptions into ±10 bands | all four neighbour breaches |
| N3 | drop `_check_the_trigonometry` from `_GUARDS` | 3 failed — the *walk* included, which is R1-M1's point: under the first revision this left the walk green |
| N4 | add a fourth `_check_*` that never joined `_GUARDS` | `test_every_walking_guard_is_registered` |
| N5 | drop `.claude` from `_NOT_OURS` | `test_the_exclusion_of_other_agents_worktrees_holds` |
| N6 | trigonometry message drops its test name, keeps the path | `test_a_guard_failure_names_the_guard_and_what_it_refuses` (R1-N1) |
| N7 | name-guard message stops saying the match is case-insensitive | `test_the_name_guard_says_it_matches_regardless_of_case` (R1-N2) |
| N8 | range rendered `{:g}` again | `test_a_guard_failure_names_the_guard_and_what_it_refuses` (R1-N3) |

## Gate

`python scripts/gate.py`, from this worktree at `b59bfec`, verbatim:

```
  whiteout source: /home/user/HTN_WHITEOUT/.claude/worktrees/agent-a101ff4b2f95b1987/whiteout/__init__.py
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (3.5s)
GATE PASS: build (0.5s)
GATE PASS: smoke (0.4s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.8s
```

`python -m pytest -q -m "not slow"` → `340 passed in 3.22s`;
`tests/test_geo.py` → `41 passed`.

`main` is currently red on `whiteout/belief/geometry.py` (merged in PR #79,
P0 #85, being fixed in parallel). `main` is **not** merged into this branch,
so that failure is not inherited here and is not caused by this PR.

## Follow-ups

Known limitations of this change, and things it deliberately did not touch:

- **PR #78's R2-N1 (unit rescaling) is untouched and still open**, and it is
  now the open item that matters most here: a precomputed cosine defeats the
  trigonometry guard, which is exactly what makes this PR's exemption ~54 m
  wide rather than sub-metre. `_SCALE = 1113.2 * 100.0` with a precomputed
  cosine still passes all three guards and the whole gate. Not planted as a
  breach, because it would not go red.
- **R2-M1 is still open**: `whiteout/geo.py`'s module docstring and `SPEC.md`
  §5 still say "Two files are exempt and no others", where
  `_TRIGONOMETRY_ALLOWED` has three entries. Not edited here — one concern per
  PR, and `SPEC.md` is not mine to touch.

---


Ollienel777:
cold: findings @ 50b60f0

1 Critical, 2 Medium, 3 Nits. The central claim holds — I re-derived it and it
is right, so the decision to take option 2 over option 1 is correct. What is
wrong is the **stated cost** of the exemption, in two places, and that is C1.

Gate from this worktree at `50b60f0`, verbatim:

```
  whiteout source: <not installed>
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (3.5s)
GATE PASS: build (0.5s)
GATE PASS: smoke (0.5s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.9s
```

`python -m pytest -q -m "not slow"` → `334 passed in 3.25s`;
`tests/test_geo.py` → `35 passed`. `whiteout.__file__` pinned to this worktree.

## What I verified, and it holds

**The rejection of option 1 is sound, to the metre.** Against `whiteout.geo`'s
own `WGS84_A`/`WGS84_E2`: equatorial circumference `2*pi*a` = **40 075 016.6856 m**,
meridian quadrant by Simpson over `M(phi)` = **10 001 965.7293 m**, full
meridian = **40 007 862.9173 m**. All three planted literals
(`tests/test_geo.py:406`, `:412`, `:433`) are the real figures. I built the
converter the PR says option 1 would free —
`LON_M_PER_DEG = 40075016.686 / 360`, `LAT_M_PER_DEG = 40007862.917 / 360` —
and under the shipped range it fails `_check_the_ellipsoid`; under
`(1e5, 9.99e6)` it passes **all three guards**, and disagrees with
`whiteout.geo` by **51.04 m** at the arena corner. The argument is not
hand-waving: option 1 really does free a working second converter.

**All eight mutations reproduce red**, on the tests named, with
`PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`:

| # | result |
|---|---|
| M1 | `1 failed` — `test_a_mavlink_adapter_may_spell_the_dege7_scaling` |
| M2 | `2 failed` — `[a neighbour of the degE7 exemption…]`, `[the quadrant of the meridian…]` |
| M3 | `4 failed` — `[the equatorial circumference…]`, `[the polar meridian…]`, plus the two above |
| M4 | `1 failed` — `test_every_guard_owns_a_planted_breach` |
| M5 | `1 failed` — `test_every_guard_owns_a_planted_breach` |
| M6 | `1 failed` — `[a conversion named by an annotated assignment]` |
| M7 | `1 failed` — `test_a_guard_failure_names_the_guard_and_what_it_refuses` |
| M8 | `1 failed` — same |

**The refactor is faithful.** `_check_*(where, text)` at `tests/test_geo.py:254`,
`:284` and `:298` are line-for-line the old test bodies; the only behavioural
change is the trigonometry guard's `continue`-per-file becoming an early
`return`, which is equivalent. `_walk` (`:340-342`) passes exactly
`(rel_path, source_text)`, and the planted breaches call
`guard("whiteout/transport/probe.py", source)` — **the same function, the same
signature, no second code path**. The walked file set is unchanged but for
`.claude`.

**The 13 planted breaches are sound and none is vacuous.** I ran every breach
against every guard. Each fails the guard it names. Exactly one also trips a
neighbour — `"WGS84_A = 6378137.0"` (`:402`) also fails the name guard, because
`WGS84_A` contains `wgs84` — and that is harmless, since the parametrised test
asserts the *targeted* guard raises. `10000001.0` (`:428`) is genuinely still a
breach: the exemption at `:263` is `value == _DEGE7`, an exact compare.

**The failure messages are accurate.** I printed all four. `[100000, 1e+08]`
matches the inclusive `low <= value <= high` at `:265`; `f, 1/f and e^2, to a
relative 1e-6` matches `_ELLIPSOID_SHAPES` and `rel_tol=1e-6` at `:276`; the
word list in the name message matches `_CONVERSION_WORDS`. The PR body's
verbatim quote is byte-correct.

**`.claude` is 27 files.** `find` over the main checkout: 101 → **27** (the
author measured 106 → 27; the worktree count drifts, the direction and the 27
are right). `git ls-files -- '.claude'` is 10 files, **none of them `.py`**,
and 27 is the whole tracked `*.py` count — so the exclusion drops nothing of
ours today.

---

## R1-C1 (Critical) — the diff misstates what the exemption costs, twice

`tests/test_geo.py:180-187`, repeated in the PR body's *"What the exemption
costs"*. Both halves of that sentence are wrong, and it is the paragraph a
future reviewer will read when deciding whether to widen the exemption.

**(a) "0.02% low, about 0.2 m per km North" is measured against the wrong
thing.** `1e7 / 90` = 111 111.111 m/deg. That *is* 0.0197% below the quadrant's
mean degree (111 132.953). But the guard exists to stop a converter disagreeing
with **`whiteout/geo.py`**, and at the strait `radii_of_curvature(71.99)` gives
`M` = 6 393 414.135, so the true latitude degree there is **111 586.127 m/deg**.
`1e7/90` is **0.4257% low — 4.26 m per km North, and 53.44 m at the arena's
12.5 km half-length.** That is **22x** the figure in the comment. This file's
own docstring (`tests/test_geo.py:13-17`) sets the rule that "everything
numeric here is evaluated at Bellot Strait's latitude"; this number is the one
that is not.

**(b) "The longitude half of any such converter still needs a cosine, which
[the trigonometry guard] refuses" is false.** It needs a cosine *value*, not a
cosine *call*, and the guard only matches `ast.Attribute` and
`from math/numpy import` (`:311`, `:319`). I wrote the whole two-axis
converter:

```python
DEGE7 = 1e7
LAT_M_PER_DEG = DEGE7 / 90.0
COS_ORIGIN = 0.3093051
def offset(lat, lon, east, north):
    return (lat + north / LAT_M_PER_DEG, lon + east / (LAT_M_PER_DEG * COS_ORIGIN))
```

`_check_the_ellipsoid` **PASS**, `_check_a_frame_conversion_name` **PASS**,
`_check_the_trigonometry` **PASS**. Worst disagreement with `whiteout.geo` over
the arena: **53.63 m** — i.e. *within 3 m of the 51.04 m converter this PR
rejects option 1 in order to prevent*. The two options are not 0.2 m against a
decade; they are ~53 m through one door against ~51 m through a decade of
doors. Option 2 is still the right call — **one door beats ten** — but the
diff's arithmetic is not the reason, and as written it would let the next
reviewer widen the exemption believing the cost is sub-metre.

A third sentence is made stale by the same thing: `tests/test_geo.py:161-162`,
*"A computed spelling does not help either: `6_378_000.0 + 137.0` is caught by
its first term."* That reasoning no longer covers the exempted term.
`METRES_PER_DEGREE = 1e7 / 90.0` is a metres-per-degree factor the guard cannot
see, in a repo where the naive `111320.0` is planted as a breach at `:403` —
111 320 is caught, 111 111 is not, and they are 0.19% apart.

**What I am asking for is text, not a redesign.** Correct the two claims at
`:180-187` and in the PR body to the strait-evaluated figures (4.26 m per km,
~53 m at the corner), and drop or qualify the cosine sentence, noting that
#78's R2-N1 precomputed-cosine bypass covers this shape too. If it is cheap,
also plant `SCALE = 9999999.0` so the *lower* neighbour of the exemption is
pinned as well as the upper one. I would not hold the merge for a code change
here; I would hold it for the two sentences, because a wrong number in a
load-bearing comment is how the next widening gets justified.

## R1-M1 (Medium) — `_GUARDS` does not drive the walk, so the coverage test is one-directional

`tests/test_geo.py:330-337` says `_GUARDS` is *"every static guard this section
runs over the tree"*, and `:389-394` says *"a guard added later cannot land
without one"*. Neither is quite true. The three walking tests at `:345-357`
each name their guard directly — `_walk(_check_the_ellipsoid)` and so on — and
never read `_GUARDS`. My M4 run confirms it: with `_check_the_trigonometry`
removed from the tuple, the trigonometry *walk* stays green and only
`test_every_guard_owns_a_planted_breach` goes red (`1 failed, 34 passed`).

So the tuple catches *removal* from `_GUARDS` but not *omission from it*: a
fourth guard added as a walking test and not registered lands with no planted
breach and no coverage in `test_a_mavlink_adapter_may_spell_the_dege7_scaling`,
silently — which is the exact hole `test_every_guard_owns_a_planted_breach` was
added to close. Cheapest fix that keeps the failure messages' embedded test
names valid: have each walking test assert its guard is in `_GUARDS`, or
collect the module's `_check_*` functions and assert that set equals `_GUARDS`.

## R1-M2 (Medium) — the `.claude` exclusion is untested, and the stated reason for that is wrong

`tests/test_geo.py:108-124` and the PR body: *"A test would have to create a
file under `.claude/` in the checkout it runs in."* It would not. The filter is
a pure predicate over `path.relative_to(REPO_ROOT).parts`; only its being
inline inside `_sources()` (`:138-143`) makes it look untestable. Lift it and
it pins with no filesystem at all — I ran exactly this:

```python
def _is_ours(rel: str) -> bool:
    return not (set(Path(rel).parts) & _NOT_OURS)

assert not _is_ours(".claude/worktrees/agent-x/whiteout/geo.py")
assert _is_ours("whiteout/geo.py") and _is_ours("scripts/gate.py")
```

This PR just lifted three guards into pure functions *for precisely this
reason* ("the guards can now be run against source that is not on disk") and
then did not do it for the one change that **removes** files from the guard's
reach. Risk today is nil — zero tracked `.py` under `.claude`, and the
exclusion is correct. Risk later is small but real: `.claude/` is where hooks
live, a hook is plausibly Python, and `.claude/**` is already outside every
ruff, mypy and pytest glob, so a converter that landed there would be unguarded
by everything at once. Untested is acceptable *given* that it is three lines
not to be.

## Nits

- **R1-N1** `tests/test_geo.py:487-493` — only the ellipsoid message is fully
  pinned. For the other two guards the test asserts only that
  `"tests/test_geo.py::"` appears, so an M7-shaped regression on those two
  (dropping the test name, keeping the path) passes. M7 and M8 were run against
  the ellipsoid guard only.
- **R1-N2** `tests/test_geo.py:289-295` — the name guard matches on
  `name.lower()` (`:287`) but the message does not say the match is
  case-insensitive, so a reader who sees `Geodetic` rejected has to open this
  file to learn why.
- **R1-N3** `tests/test_geo.py:163` and `:268` — `{low:g}` renders the bracket
  as `[100000, 1e+08]`, two spellings of one kind of number. `{low:.0e}` reads
  as `[1e+05, 1e+08]` and matches the constant as written.

---

## Ruling: the `1e6` follow-up **should be pre-empted in this PR**

I went in expecting to say "one concern per PR", and the numbers changed my
mind. Three things decided it.

**1. The marginal hole is provably zero.** `1e6 / 9 = 111 111.111` and
`1e7 / 90 = 111 111.111` — bit-identical. Every derived metres-per-degree a
`1e6` exemption could reach, the `1e7` exemption already reaches. And unlike
`1e7`, which sits 2 km from the meridian quadrant, **nothing geodetic lives
near 1e6**: `1e6/90` = 11 111 (ten times too small), `1e6/360` = 2 778; no
radius, no circumference, no quadrant. "An exemption nothing needs is an
exemption nobody can justify later" is the right instinct, but the
justification here is checkable against the MAVLink message definition rather
than against this repo's state — `time_usec` is microseconds, the same class of
fact as `GLOBAL_POSITION_INT` being degE7 — and it costs strictly less than the
exemption already being shipped.

**2. It is not a second concern.** It is a second entry in a tuple this PR
creates, plus a comment line and a planted neighbour. The concern is one
sentence long: *the metre-scale range must not adjudicate MAVLink wire
factors.* Splitting that across two PRs buys a reviewer nothing.

**3. Leaving it recreates the exact condition #82 was filed to remove.** #82
exists because #78 recorded a known future-blocking guard hole and shipped. The
PR body now records the identical hole and ships. The cost when #64 hits it is
not the two-minute edit — the new message does make that edit obvious — it is
that a transport ticket has to open and defend a change to `tests/test_geo.py`,
a file in nobody's area, mid-event, with about four hours to freeze.

The honest argument on the other side, which I checked: `1e6` is **not strictly
blocking**, because `message.time_usec * 1e-6` passes the guard today (1e-6 is
below the range), as does `* 1e-7` for position. So #64 has an escape hatch. I
do not think that saves it — pushing an adapter author into `* 1e-6` to appease
a geodesy guard is worse code and a worse error to have hit — but it does mean
this is a judgement call rather than a defect, which is why it is a ruling here
and not a finding.

**Concretely:** hold both in one place (`_WIRE_FACTORS = (1.0e7, 1.0e6)`, or
keep `_DEGE7` and add `_USEC`), say in the comment that these are dimensionless
MAVLink wire factors and not metres, plant `SCALE = 1000001.0` as a neighbour
breach, and keep the message's "name it beside that one" sentence — it is good,
and it should now read as naming a list.

---

Ollienel777 (commented):
cold: findings @ 50b60f0

1 Critical, 2 Medium, 3 Nits. The central claim holds — I re-derived it and it
is right, so the decision to take option 2 over option 1 is correct. What is
wrong is the **stated cost** of the exemption, in two places, and that is C1.

Gate from this worktree at `50b60f0`, verbatim:

```
  whiteout source: <not installed>
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (3.5s)
GATE PASS: build (0.5s)
GATE PASS: smoke (0.5s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.9s
```

`python -m pytest -q -m "not slow"` → `334 passed in 3.25s`;
`tests/test_geo.py` → `35 passed`. `whiteout.__file__` pinned to this worktree.

## What I verified, and it holds

**The rejection of option 1 is sound, to the metre.** Against `whiteout.geo`'s
own `WGS84_A`/`WGS84_E2`: equatorial circumference `2*pi*a` = **40 075 016.6856 m**,
meridian quadrant by Simpson over `M(phi)` = **10 001 965.7293 m**, full
meridian = **40 007 862.9173 m**. All three planted literals
(`tests/test_geo.py:406`, `:412`, `:433`) are the real figures. I built the
converter the PR says option 1 would free —
`LON_M_PER_DEG = 40075016.686 / 360`, `LAT_M_PER_DEG = 40007862.917 / 360` —
and under the shipped range it fails `_check_the_ellipsoid`; under
`(1e5, 9.99e6)` it passes **all three guards**, and disagrees with
`whiteout.geo` by **51.04 m** at the arena corner. The argument is not
hand-waving: option 1 really does free a working second converter.

**All eight mutations reproduce red**, on the tests named, with
`PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`:

| # | result |
|---|---|
| M1 | `1 failed` — `test_a_mavlink_adapter_may_spell_the_dege7_scaling` |
| M2 | `2 failed` — `[a neighbour of the degE7 exemption…]`, `[the quadrant of the meridian…]` |
| M3 | `4 failed` — `[the equatorial circumference…]`, `[the polar meridian…]`, plus the two above |
| M4 | `1 failed` — `test_every_guard_owns_a_planted_breach` |
| M5 | `1 failed` — `test_every_guard_owns_a_planted_breach` |
| M6 | `1 failed` — `[a conversion named by an annotated assignment]` |
| M7 | `1 failed` — `test_a_guard_failure_names_the_guard_and_what_it_refuses` |
| M8 | `1 failed` — same |

**The refactor is faithful.** `_check_*(where, text)` at `tests/test_geo.py:254`,
`:284` and `:298` are line-for-line the old test bodies; the only behavioural
change is the trigonometry guard's `continue`-per-file becoming an early
`return`, which is equivalent. `_walk` (`:340-342`) passes exactly
`(rel_path, source_text)`, and the planted breaches call
`guard("whiteout/transport/probe.py", source)` — **the same function, the same
signature, no second code path**. The walked file set is unchanged but for
`.claude`.

**The 13 planted breaches are sound and none is vacuous.** I ran every breach
against every guard. Each fails the guard it names. Exactly one also trips a
neighbour — `"WGS84_A = 6378137.0"` (`:402`) also fails the name guard, because
`WGS84_A` contains `wgs84` — and that is harmless, since the parametrised test
asserts the *targeted* guard raises. `10000001.0` (`:428`) is genuinely still a
breach: the exemption at `:263` is `value == _DEGE7`, an exact compare.

**The failure messages are accurate.** I printed all four. `[100000, 1e+08]`
matches the inclusive `low <= value <= high` at `:265`; `f, 1/f and e^2, to a
relative 1e-6` matches `_ELLIPSOID_SHAPES` and `rel_tol=1e-6` at `:276`; the
word list in the name message matches `_CONVERSION_WORDS`. The PR body's
verbatim quote is byte-correct.

**`.claude` is 27 files.** `find` over the main checkout: 101 → **27** (the
author measured 106 → 27; the worktree count drifts, the direction and the 27
are right). `git ls-files -- '.claude'` is 10 files, **none of them `.py`**,
and 27 is the whole tracked `*.py` count — so the exclusion drops nothing of
ours today.

---

## R1-C1 (Critical) — the diff misstates what the exemption costs, twice

`tests/test_geo.py:180-187`, repeated in the PR body's *"What the exemption
costs"*. Both halves of that sentence are wrong, and it is the paragraph a
future reviewer will read when deciding whether to widen the exemption.

**(a) "0.02% low, about 0.2 m per km North" is measured against the wrong
thing.** `1e7 / 90` = 111 111.111 m/deg. That *is* 0.0197% below the quadrant's
mean degree (111 132.953). But the guard exists to stop a converter disagreeing
with **`whiteout/geo.py`**, and at the strait `radii_of_curvature(71.99)` gives
`M` = 6 393 414.135, so the true latitude degree there is **111 586.127 m/deg**.
`1e7/90` is **0.4257% low — 4.26 m per km North, and 53.44 m at the arena's
12.5 km half-length.** That is **22x** the figure in the comment. This file's
own docstring (`tests/test_geo.py:13-17`) sets the rule that "everything
numeric here is evaluated at Bellot Strait's latitude"; this number is the one
that is not.

**(b) "The longitude half of any such converter still needs a cosine, which
[the trigonometry guard] refuses" is false.** It needs a cosine *value*, not a
cosine *call*, and the guard only matches `ast.Attribute` and
`from math/numpy import` (`:311`, `:319`). I wrote the whole two-axis
converter:

```python
DEGE7 = 1e7
LAT_M_PER_DEG = DEGE7 / 90.0
COS_ORIGIN = 0.3093051
def offset(lat, lon, east, north):
    return (lat + north / LAT_M_PER_DEG, lon + east / (LAT_M_PER_DEG * COS_ORIGIN))
```

`_check_the_ellipsoid` **PASS**, `_check_a_frame_conversion_name` **PASS**,
`_check_the_trigonometry` **PASS**. Worst disagreement with `whiteout.geo` over
the arena: **53.63 m** — i.e. *within 3 m of the 51.04 m converter this PR
rejects option 1 in order to prevent*. The two options are not 0.2 m against a
decade; they are ~53 m through one door against ~51 m through a decade of
doors. Option 2 is still the right call — **one door beats ten** — but the
diff's arithmetic is not the reason, and as written it would let the next
reviewer widen the exemption believing the cost is sub-metre.

A third sentence is made stale by the same thing: `tests/test_geo.py:161-162`,
*"A computed spelling does not help either: `6_378_000.0 + 137.0` is caught by
its first term."* That reasoning no longer covers the exempted term.
`METRES_PER_DEGREE = 1e7 / 90.0` is a metres-per-degree factor the guard cannot
see, in a repo where the naive `111320.0` is planted as a breach at `:403` —
111 320 is caught, 111 111 is not, and they are 0.19% apart.

**What I am asking for is text, not a redesign.** Correct the two claims at
`:180-187` and in the PR body to the strait-evaluated figures (4.26 m per km,
~53 m at the corner), and drop or qualify the cosine sentence, noting that
#78's R2-N1 precomputed-cosine bypass covers this shape too. If it is cheap,
also plant `SCALE = 9999999.0` so the *lower* neighbour of the exemption is
pinned as well as the upper one. I would not hold the merge for a code change
here; I would hold it for the two sentences, because a wrong number in a
load-bearing comment is how the next widening gets justified.

## R1-M1 (Medium) — `_GUARDS` does not drive the walk, so the coverage test is one-directional

`tests/test_geo.py:330-337` says `_GUARDS` is *"every static guard this section
runs over the tree"*, and `:389-394` says *"a guard added later cannot land
without one"*. Neither is quite true. The three walking tests at `:345-357`
each name their guard directly — `_walk(_check_the_ellipsoid)` and so on — and
never read `_GUARDS`. My M4 run confirms it: with `_check_the_trigonometry`
removed from the tuple, the trigonometry *walk* stays green and only
`test_every_guard_owns_a_planted_breach` goes red (`1 failed, 34 passed`).

So the tuple catches *removal* from `_GUARDS` but not *omission from it*: a
fourth guard added as a walking test and not registered lands with no planted
breach and no coverage in `test_a_mavlink_adapter_may_spell_the_dege7_scaling`,
silently — which is the exact hole `test_every_guard_owns_a_planted_breach` was
added to close. Cheapest fix that keeps the failure messages' embedded test
names valid: have each walking test assert its guard is in `_GUARDS`, or
collect the module's `_check_*` functions and assert that set equals `_GUARDS`.

## R1-M2 (Medium) — the `.claude` exclusion is untested, and the stated reason for that is wrong

`tests/test_geo.py:108-124` and the PR body: *"A test would have to create a
file under `.claude/` in the checkout it runs in."* It would not. The filter is
a pure predicate over `path.relative_to(REPO_ROOT).parts`; only its being
inline inside `_sources()` (`:138-143`) makes it look untestable. Lift it and
it pins with no filesystem at all — I ran exactly this:

```python
def _is_ours(rel: str) -> bool:
    return not (set(Path(rel).parts) & _NOT_OURS)

assert not _is_ours(".claude/worktrees/agent-x/whiteout/geo.py")
assert _is_ours("whiteout/geo.py") and _is_ours("scripts/gate.py")
```

This PR just lifted three guards into pure functions *for precisely this
reason* ("the guards can now be run against source that is not on disk") and
then did not do it for the one change that **removes** files from the guard's
reach. Risk today is nil — zero tracked `.py` under `.claude`, and the
exclusion is correct. Risk later is small but real: `.claude/` is where hooks
live, a hook is plausibly Python, and `.claude/**` is already outside every
ruff, mypy and pytest glob, so a converter that landed there would be unguarded
by everything at once. Untested is acceptable *given* that it is three lines
not to be.

## Nits

- **R1-N1** `tests/test_geo.py:487-493` — only the ellipsoid message is fully
  pinned. For the other two guards the test asserts only that
  `"tests/test_geo.py::"` appears, so an M7-shaped regression on those two
  (dropping the test name, keeping the path) passes. M7 and M8 were run against
  the ellipsoid guard only.
- **R1-N2** `tests/test_geo.py:289-295` — the name guard matches on
  `name.lower()` (`:287`) but the message does not say the match is
  case-insensitive, so a reader who sees `Geodetic` rejected has to open this
  file to learn why.
- **R1-N3** `tests/test_geo.py:163` and `:268` — `{low:g}` renders the bracket
  as `[100000, 1e+08]`, two spellings of one kind of number. `{low:.0e}` reads
  as `[1e+05, 1e+08]` and matches the constant as written.

---

## Ruling: the `1e6` follow-up **should be pre-empted in this PR**

I went in expecting to say "one concern per PR", and the numbers changed my
mind. Three things decided it.

**1. The marginal hole is provably zero.** `1e6 / 9 = 111 111.111` and
`1e7 / 90 = 111 111.111` — bit-identical. Every derived metres-per-degree a
`1e6` exemption could reach, the `1e7` exemption already reaches. And unlike
`1e7`, which sits 2 km from the meridian quadrant, **nothing geodetic lives
near 1e6**: `1e6/90` = 11 111 (ten times too small), `1e6/360` = 2 778; no
radius, no circumference, no quadrant. "An exemption nothing needs is an
exemption nobody can justify later" is the right instinct, but the
justification here is checkable against the MAVLink message definition rather
than against this repo's state — `time_usec` is microseconds, the same class of
fact as `GLOBAL_POSITION_INT` being degE7 — and it costs strictly less than the
exemption already being shipped.

**2. It is not a second concern.** It is a second entry in a tuple this PR
creates, plus a comment line and a planted neighbour. The concern is one
sentence long: *the metre-scale range must not adjudicate MAVLink wire
factors.* Splitting that across two PRs buys a reviewer nothing.

**3. Leaving it recreates the exact condition #82 was filed to remove.** #82
exists because #78 recorded a known future-blocking guard hole and shipped. The
PR body now records the identical hole and ships. The cost when #64 hits it is
not the two-minute edit — the new message does make that edit obvious — it is
that a transport ticket has to open and defend a change to `tests/test_geo.py`,
a file in nobody's area, mid-event, with about four hours to freeze.

The honest argument on the other side, which I checked: `1e6` is **not strictly
blocking**, because `message.time_usec * 1e-6` passes the guard today (1e-6 is
below the range), as does `* 1e-7` for position. So #64 has an escape hatch. I
do not think that saves it — pushing an adapter author into `* 1e-6` to appease
a geodesy guard is worse code and a worse error to have hit — but it does mean
this is a judgement call rather than a defect, which is why it is a ruling here
and not a finding.

**Concretely:** hold both in one place (`_WIRE_FACTORS = (1.0e7, 1.0e6)`, or
keep `_DEGE7` and add `_USEC`), say in the comment that these are dimensionless
MAVLink wire factors and not metres, plant `SCALE = 1000001.0` as a neighbour
breach, and keep the message's "name it beside that one" sentence — it is good,
and it should now read as naming a list.

---

Ollienel777:
loop: fixer dispatched @ 50b60f0 — findings

---

Ollienel777:
fix: pushed @ b59bfec

**R1-C1 (Critical) — taken, both halves, and a third number found wrong while checking.**
Measured at the strait against `whiteout/geo.py`, not the global mean:
`radii_of_curvature(71.99)` gives M = 6 393 414.135, i.e. **111 586.127 m/deg**,
so `1e7/90` = 111 111.111 is **0.426% low — about 4.3 m per km North and
53.44 m at the arena's 12.5 km half-length**. The containment claim is gone:
`COS_ORIGIN = 0.3093051` is neither an `ast.Attribute` nor a `from math import`,
and the full two-axis converter passes all three guards with **53.63 m** worst
disagreement over the arena — I ran the three guards over it and printed
`PASS, PASS, PASS`. The comment at the old `:180-187` now says outright that
this exemption opens a ~54 m two-axis converter and that none of the three
guards catches it. The stale "computed spelling is caught by its first term"
sentence now says it is caught *unless the first term is an exempted wire
factor*. While measuring, the neighbouring "`111320.0` disagrees by about
90 m over the strait" turned out to be wrong too — it is **30.0 m** — so that
is corrected, and the comment states the resulting fact plainly: the guard
refuses the less wrong spelling (30.0 m, caught) and admits the more wrong one
(53.63 m, exempt). Also planted `9999999.0`, so both neighbours are pinned.

**The `1e6` ruling — implemented.** `_WIRE_FACTORS` is a dict of the two
factors to what each decodes (`1e7: degE7 lat/lon`, `1e6: time_usec
microseconds`), so a third entry cannot be added without naming its wire
field and the failure message derives the whole list rather than drifting from
it. The comment carries the one sentence of concern and the reason the
marginal hole is zero (`1e6/9` and `1e7/90` are the same float; `1e6/90` is
11 111 and `1e6/360` is 2 778). Planted `1000001.0` and `999999.0`, and the
adapter fixture now spells `1e6` as well. Guard message now reads
`Its only exemptions are 1e+07 (degE7 lat/lon), 1e+06 (time_usec microseconds).`

**R1-M1 (Medium) — taken.** `_walk` asserts its guard is in `_GUARDS`, and a
new `test_every_walking_guard_is_registered` collects the module's `_check_*`
functions and fails any that never joined the tuple. Dropping
`_check_the_trigonometry` from `_GUARDS` now takes the *walk* red as well,
which is exactly what M4 showed it did not.

**R1-M2 (Medium) — taken, and the stated reason withdrawn.** The filter is
lifted into `_is_ours(rel_path)`, a pure predicate over the path's parts, and
`test_the_exclusion_of_other_agents_worktrees_holds` pins it with no
filesystem — `.claude/worktrees/...`, `.claude/hooks/...`, `.egg-info`,
`.venv`, plus the three that must stay ours and two that must not be caught by
prefix. The "a test would have to create a file under `.claude/`" line is gone
from the PR body.

**Nits — all three taken.** N1: the name and trigonometry guards are now
pinned on their full `tests/test_geo.py::<test_name>`, not just the path.
N2: the name-guard message says the match is "matched without regard to case",
with a test on `GeodeticFix`. N3: the range renders `[1e+05, 1e+08]`.

**PR body** rewritten for all of the above and read back.

**Mutations run** (`PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`,
`tests/test_geo.py`, 41 tests), each red on the named test:

| # | mutation | went red |
|---|---|---|
| N1 | drop the `1e6` wire factor | `test_a_mavlink_adapter_may_spell_the_wire_factors` |
| N2 | widen the exemptions into ±10 bands | all four neighbour breaches |
| N3 | drop `_check_the_trigonometry` from `_GUARDS` | `test_nothing_but_whiteout_geo_does_trigonometry_on_a_latitude`, `test_every_walking_guard_is_registered`, `test_every_guard_owns_a_planted_breach` |
| N4 | a fourth `_check_*` that never joined `_GUARDS` | `test_every_walking_guard_is_registered` |
| N5 | drop `.claude` from `_NOT_OURS` | `test_the_exclusion_of_other_agents_worktrees_holds` |
| N6 | trigonometry message drops its test name, keeps the path | `test_a_guard_failure_names_the_guard_and_what_it_refuses` |
| N7 | name-guard message stops saying the match is case-insensitive | `test_the_name_guard_says_it_matches_regardless_of_case` |
| N8 | range rendered `{:g}` again | `test_a_guard_failure_names_the_guard_and_what_it_refuses` |

**Gate**, `python scripts/gate.py` from this worktree at `b59bfec`, verbatim:

```
  whiteout source: /home/user/HTN_WHITEOUT/.claude/worktrees/agent-a101ff4b2f95b1987/whiteout/__init__.py
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (3.5s)
GATE PASS: build (0.5s)
GATE PASS: smoke (0.4s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 4.8s
```

`python -m pytest -q -m "not slow"` → `340 passed in 3.22s`;
`tests/test_geo.py` → `41 passed`. `main` was not merged in, so #85's
`whiteout/belief/geometry.py` failures are not inherited here.


---

Ollienel777:
semi-cold: closes @ b59bfec

---

Ollienel777 (commented):
semi-cold: closes @ b59bfec

6 of 6 closed. 5 new nits, none holding. I re-derived every figure against
`whiteout.geo` rather than against the fixer's report, and re-ran six of the
eight mutations plus three of my own. One number in the rewritten C1 paragraph
is still off — in the *understating* direction — but by 6%, not 22x, and it
does not touch the decision. That is R2-N1.

Gate from this worktree at `b59bfec`, verbatim:

```
  whiteout source: /home/user/HTN_WHITEOUT/.claude/worktrees/agent-a312b759169d6f7e0/whiteout/__init__.py
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.1s)
GATE PASS: test (3.6s)
GATE PASS: build (0.4s)
GATE PASS: smoke (0.4s)
GATE PASS: determinism (0.2s)
GATE OK: 8 steps in 6.8s
```

`python -m pytest -q -m "not slow"` -> `340 passed in 3.22s`;
`tests/test_geo.py` -> `41 passed`. `main` not merged in.

---

## Verdicts

**R1-C1 (Critical) — closed.** Both halves re-derived independently, and they
are right to the metre:

| claim | mine |
|---|---|
| `M(71.99)` | `6393414.13541103` |
| latitude degree at the strait | `111586.12710646905` m/deg |
| `1e7 / 90` | `111111.11111111111` |
| low by | **0.4257%** of the true degree |
| at 12.5 km North | `12500 * (111586.127/111111.111 - 1)` = **53.439 m** |
| per km North | **4.28 m** |

So 4.3 m/km and 53.44 m are correct, and the old 0.2 m/km was measured against
the global mean. The containment claim is gone and its replacement is true: I
pasted the comment's own converter into `_check_the_ellipsoid`,
`_check_a_frame_conversion_name` and `_check_the_trigonometry` and got
`PASS PASS PASS`. The "caught by its first term" sentence is now correctly
qualified.

**The unprompted correction is right, and so is the conclusion it draws.** The
naive `111320.0` spelling: **29.883 m** north-only at 12.5 km, **29.98 m**
worst over the 25 km x 2 km arena measured through `geodetic_to_local` — so
30.0 m, not 90 m. The uncomfortable sentence holds: 111 320 is a planted
breach and 111 111 is exempt, and 111 111 is **1.79x further** from
`whiteout.geo`. Keep that sentence.

One figure in the same paragraph does not reproduce — R2-N1 below.

**The `1e6` pre-emption — closed.** `_WIRE_FACTORS` is a dict, and I checked
it opens nothing a tuple would not: membership is `abs(float(...)) in dict`,
which is exact float equality by hash, identical to `in (1e7, 1e6)`. The
naming is not decorative — `_WIRE_FACTORS_NAMED` is derived at import, and
`test_a_guard_failure_names_the_guard_and_what_it_refuses` now loops the dict
asserting both `{value:.0e}` *and* the reason string appear, so a third entry
cannot land unnamed and cannot drift. `1e6 / 9` and `1e7 / 90` are
bit-identical at the hex level — both `0x1.b2071c71c71c7p+16` — so the
marginal hole really is zero. `1e6/90` = 11 111 and `1e6/360` = 2 778 check
out; nothing geodetic is near either. `10000001.0`, `9999999.0`, `1000001.0`
and `999999.0` are all planted and all refused.

**R1-M1 (Medium) — closed.** This is the one I most wanted to see fail, and it
does not. Dropping `_check_the_trigonometry` from `_GUARDS` (N3):

```
FAILED tests/test_geo.py::test_nothing_but_whiteout_geo_does_trigonometry_on_a_latitude
FAILED tests/test_geo.py::test_every_walking_guard_is_registered
FAILED tests/test_geo.py::test_every_guard_owns_a_planted_breach
3 failed, 38 passed
```

The **walk** is red, which is exactly what round 1's M4 showed it was not.
N4 (a fourth `_check_*` that never joins the tuple) is red on
`test_every_walking_guard_is_registered` alone. Both directions are now
covered. Residual hole is mine, R2-N3.

**R1-M2 (Medium) — closed.** `_is_ours(rel_path)` is a pure predicate and the
new test pins it with no filesystem; N5 (drop `.claude` from `_NOT_OURS`) goes
red on it. The wrong reason is gone from the PR body. Incidental improvement
worth noting: the `.egg-info` half now reads the *relative* parts where
`_sources()` read the absolute ones, which is a tightening in the right
direction. Residual is R2-N2.

**R1-N1 — closed.** Both remaining guards are now pinned on the full
`tests/test_geo.py::<test_name>`, not just the path. N8 and my own message
mutations confirm. One message site is still unexercised — R2-N4.

**R1-N2 — closed.** Message says "matched without regard to case" and
`test_the_name_guard_says_it_matches_regardless_of_case` pins it; deleting
just that phrase (keeping the test name) goes red on it.

**R1-N3 — closed.** `[1e+05, 1e+08]` in both the constant and the message; N8
(`{:g}` again) goes red.

**Mutations re-run** — six of eight, plus three of mine, with
`PYTHONDONTWRITEBYTECODE=1 -p no:cacheprovider`:

| # | result |
|---|---|
| N1 | `1 failed` — `test_a_mavlink_adapter_may_spell_the_wire_factors` |
| N3 | `3 failed` — **the walk included** |
| N4 | `1 failed` — `test_every_walking_guard_is_registered` |
| N5 | `1 failed` — `test_the_exclusion_of_other_agents_worktrees_holds` |
| N6 (ImportFrom message) | `1 failed` — `test_a_guard_failure_names_the_guard_and_what_it_refuses` |
| N6 (Attribute message) | **`41 passed`** — see R2-N4 |
| N8 | `1 failed` — `test_a_guard_failure_names_the_guard_and_what_it_refuses` |
| X1 (mine) `_ellipsoid_check` defined, not walked | `41 passed` |
| X2 (mine) `_fourth_rule` walked without `_walk` | **`42 passed`** — see R2-N3 |
| X3 (mine) drop only the word "case" from the name message | `1 failed` |

**Attacks that did not get through.** Ten evasions run against the post-fix
guards: `1e6 * 10.0 / 90.0`, the int literal `10000000`, `1_000_000.0 / 9.0`,
a negated `-1e7`, `1e7` buried in a nested expression, a `1e7` reused as a
radius — all of these are the exemption doing what it says on the tin, and all
are already stated in the comment. Three that get through and are **not** this
PR's doing, for the record rather than as findings: `1e7j` (the guard tests
`isinstance(node.value, int | float)`, so a complex literal is skipped),
`Decimal("6378137.0")` and a constant read out of a JSON string — all
pre-existing, and none is a shape anybody writes by accident.

---

## New findings

### R2-N1 (Nit) — the C1 paragraph's worked constant does not produce its own number

`tests/test_geo.py:212-221`, repeated in the PR body. The comment names a
specific constant and a specific result:

> A precomputed `COS_ORIGIN = 0.3093051` ... passes all three guards, with a
> worst disagreement of **53.63 m** over the arena.

With that constant literally as written, I measure **56.729 m** at
`(-12500, -1000)`, not 53.63 m. 53.63 m is what you get only if the cosine is
*exactly* right — the true ratio is `lon_m_per_deg / lat_m_per_deg` =
`0.309382173`, which gives **53.610 m**, because both axes then carry the same
0.4257%. `0.3093051` is `cos(71.98264)`, which is nothing in particular, and
the constant a real author would actually write, `cos(71.99)` = `0.30918298`,
gives **61.675 m**.

| `COS_ORIGIN` | worst over arena |
|---|---|
| `0.3093051` (as written in the comment) | **56.729 m** |
| `0.309382173` (the exact ratio) | 53.610 m |
| `0.30918298` = `cos(71.99)` | **61.675 m** |

So the paragraph whose whole job is to stop the next person citing a too-small
cost is itself **3 to 8 m low**. It is the right direction to be wrong in only
in the sense that the real cost is bigger; the decision does not move, and
"~54 m" is still the right order. But this file's culture is that numbers are
measured, and this one is not reproducible from its own worked example.
One-line fix, either way round: say 56.7 m beside `0.3093051`, or drop the
specific constant and say the cosine is whatever makes the East axis exact, in
which case 53.6 m is right. Same for the PR body. (For what it is worth the
neighbouring 30.0 m and 51 m *are* both measured under the exact-cosine
convention — 29.98 m and 51.13 m — so the two halves of the "less wrong
spelling / more wrong spelling" comparison are apples to apples, and that
sentence needs no change.)

### R2-N2 (Nit) — `_is_ours` skips `build`, `dist`, `venv` and `node_modules` at any depth, and `docs/build/` is a real directory of ours

`tests/test_geo.py:136-144`. The predicate is a set intersection over *all* the
path's parts, so the name matches at any depth, not just at the root:

```
_is_ours("docs/build/x.py")           = False
_is_ours("whiteout/build/planner.py") = False
_is_ours("whiteout/dist/x.py")        = False
_is_ours("scripts/venv/x.py")         = False
```

`docs/build/` exists and is tracked (17 files), and it is the build loop's own
brief. No `.py` under it today, so nothing is unguarded right now — but this is
precisely the strict-direction failure the new test's own docstring says it is
there to prevent, and the test does not pin it: it pins the prefix case
(`whiteout/build_plan.py`, `scripts/claude_report.py`) and the depth-0 case
(`.claude/...`, `.venv/...`), never a skipped name at depth > 0. Pre-existing
shape, not introduced here — `_sources()` had the same intersection — but
R1-M2 asked for this predicate to be pinned, and this is the half that is not.
Cheapest fix is one more assertion recording the decision either way, e.g.
`assert not _is_ours("docs/build/x.py")` with a line saying why that is
intended; the more correct fix is to anchor `build`/`dist` to `parts[0]`, which
is where the packaging output actually lives (top-level `build/` does hold a
copy of `whiteout/geo.py`, so it genuinely must stay excluded).

### R2-N3 (Nit) — `test_every_walking_guard_is_registered` is keyed on the `_check_` prefix

`tests/test_geo.py:440-453`. It collects `globals()` entries whose name starts
with `_check_`. A rule spelt any other way is collected by neither it nor
`_walk`'s new assertion, if the test that walks it loops `_sources()` itself
rather than calling `_walk`. I planted exactly that:

```python
def _fourth_rule(where: str, text: str) -> None: ...

def test_the_fourth_rule() -> None:
    for source in _sources():
        _fourth_rule(source.relative_to(REPO_ROOT).as_posix(), source.read_text("utf-8"))
```

`42 passed` — a fourth walking rule, no planted breach, no coverage, silent.
This is the irreducible residue of a convention-based registry and I am not
asking for a fix; the two doors R1-M1 named are both shut, which is what was
asked. Worth one sentence in the `_GUARDS` comment so the next author knows the
convention is load-bearing: a walking rule must be named `_check_*` and must go
through `_walk`.

### R2-N4 (Nit) — the trigonometry guard has two failure messages and only one is pinned

`tests/test_geo.py:377-393`. `_check_the_trigonometry` raises from two places:
the `ast.Attribute` branch ("... calls cos()") and the `ImportFrom` branch
("... imports radians").
`test_a_guard_failure_names_the_guard_and_what_it_refuses` exercises only the
second, with `"from math import radians\n"`. Mutating the **Attribute** message
to drop its test name and keep the path gives `41 passed`; the same mutation on
the ImportFrom message goes red. The Attribute branch is the one a real
`math.cos(lat)` hits, so it is the message more likely to be read. One extra
tuple entry with `"import math\nC = math.cos(1.2)\n"` closes it.

### R2-N5 (Nit) — two counts in the PR body are off

`_PLANTED_BREACHES` has **16** entries, not 17; the body says "17 planted
breaches" and "`test_a_guard_fails_the_source_it_exists_to_catch` (17 cases)".
13 before this commit, plus the three new neighbours. Also "It is a second
entry in a **tuple** this PR creates" — it is a dict now, which is the better
shape and the reason the ruling was taken as asked.

---

Ollienel777:
settled: @ f9c82b0

---

PR #116, opened 2026-09-20
chore(belief): record the site's parameters, and settle convergence

### LINKED ISSUE #112: chore: record the site's real parameters from /api/site, and settle whether convergence applies
The belief field's channel geometry has no cited source. #102 corrected a real
error — `DEFAULT_STRAIT` ran 25.1 km against a 6.5 km site, because `ARENA.md`
§2's "25 km" is the deck's Northwest Passage context and not the render — but
the centreline it kept is, in its own words, "inherited and unverified", and it
sits roughly 470 m off the vessel's observed track.

There is an endpoint we have never called. `GET :8090/api/site` returns the
site centre, `bounds3413` (the EPSG:3413 extent) and `convergence_deg`.

Two deliverables.

**1. Read that endpoint and record its answer as a committed fixture**, so the
site's parameters have a citation rather than a hand-copied constant. The arena
is not reachable from CI, so the code must work from the fixture and the fetch
must be an explicit opt-in.

**2. Settle `convergence_deg`.** Grid north versus true north is a heading bias
we account for nowhere. `whiteout/vision/projection.py` turns an attitude into a
ray and `whiteout/geo.py`'s `bearing_deg` turns a position pair into a heading.
If convergence affects either, fix it and say by how many metres at the arena's
scale. If it does not, write the reasoning down in the module that would have
been wrong, so the next person does not re-derive it. A clear "this does not
apply to us, here is why" is a good outcome; do not manufacture a change to have
one.

**Out of scope:** refitting the centreline against the DEM. That needs the
heightmap and the distance-to-shore field, and it must be done per site because
`competition.lock` pins three. This issue is about having the site's real
parameters on hand and knowing whether convergence matters.

Ollienel777:
Closes #112

## What

Adds `whiteout/data/site.json`, a committed record of the arena's own answers about the site it renders, with the provenance of each answer beside it; `whiteout/site.py`, which reads that record and never opens a socket; and `scripts/arena_site.py`, which records what `:8090` answers. `tests/test_site.py` fails if `geometry.py`'s three site constants and the record disagree. The record holds a real fetch from the SIM-5 arena at `10.99.4.1`, both endpoints together, with the organisers' `MAPBOX_TOKEN` redacted.

Writes down, in `whiteout/vision/projection.py` and in summary at `whiteout/geo.py`'s `bearing_deg`, what `convergence_deg` does and does not reach. **No behaviour changes and no number moves.**

## Why

The belief field's channel geometry had no cited source. #102 corrected a real error, and its own message says the centreline it kept is "inherited and unverified". The first half of fixing that is having the site's real parameters on hand with a citation; the second half, refitting the centreline, needs the DEM and is out of scope here.

`convergence_deg` is a 49.8° heading quantity we account for nowhere, which is either a serious bug or a non-issue, and nobody had decided which.

## The convergence answer, reconciled against #121

**Nothing under `whiteout/` applies it, and the open question is named rather than closed.**

What is settled: no easting, northing or scale factor of EPSG:3413 is read or written anywhere under `whiteout/`. Positions arrive geodetic and leave geodetic; the one frame between is `whiteout.geo`'s tangent plane, whose north axis is true north by construction. A rotation between frames corrects a quantity expressed in one of them, and none of ours is.

What is **not** settled, and what an earlier revision of this PR wrongly claimed was: whether `ATTITUDE.yaw` is referenced to true north or to grid north. That revision said "the simulator's world frame is an East–North–Up frame about a geodetic origin". #121's `truth_probe.py` measured Gazebo's world x/y to be EPSG:3413 grid axes — 1547.6 m of error unrotated, 5.7 m rotated. The premise is false and has been removed.

The conclusion survives on a narrower footing, stated as the inference it is: ArduPilot's EKF fuses GPS-derived position against attitude-rotated inertial data, and MAVLink lat/lon is geodetically correct (it is the reference #121 fitted Gazebo's coordinates to), so a grid-referenced attitude would put a 49.8° disagreement inside the filter and guided flight would not track. #118 flew the fleet to waypoints. #110's false alarms landing in the channel point the same way. **So no correction is applied because no measurement supports applying one, not because the question is closed** — rotating on the strength of #121's *position* finding would be the same mistake in the other direction and the larger one.

The check that would settle it is much cheaper now and needs no moving asset, which is what defeated the last attempt: Gazebo's `~/pose/info` carries each model's orientation beside the position `truth_probe` keeps, and the two hypotheses differ by exactly the convergence. A parked quadcopter answers it.

## How to verify

```
python -m pytest tests/test_site.py -q      # 54 passed
```

`test_the_geometry_constants_are_the_recorded_ones` is the one that matters: it ties `SITE_CENTRE_LAT`, `SITE_CENTRE_LON` and `SITE_EXTENT_M` to the record, with exact equality, because those literals are a transcription and a transcription is either right or it is not. `test_a_coarsely_quoted_site_centre_does_not_move_the_parameters` keeps the loose 50 m corroboration bound from feeding that strict check. `test_loading_the_record_opens_no_socket` pins the CI half of the split.

**The fetch** is opt-in twice: no endpoint on the command line or in `WHITEOUT_ARENA_ENDPOINT` and it refuses to run; no `--write` and it prints to stdout. After a successful `--write` it says whether the answer moves `geometry.py`'s constants, so whoever is holding the arena hears about it before they commit rather than from CI.

**The recorded numbers**, checked against an independent EPSG:3413 implementation: reprojecting `SITE_LAT`/`SITE_LON` by the closed form lands on the centre of the recorded `bounds3413` to **0.0 m**, the box is square to the last decimal, and the recorded convergence sits **0.0176°** from the polar stereographic identity — 1.00 m across the half-extent, and EPSG:3413 confirmed, which is the premise the projection section reasons about.

## Gate

```
GATE PASS: install · lint · format · typecheck (34 files) · test (773 passed, 12 skipped)
GATE PASS: build · smoke · determinism
GATE OK: 8 steps
```

`origin/main` is merged in, so #121 and #118 are included.

## Follow-ups

- **Refitting `DEFAULT_STRAIT`'s centreline**, which sits roughly 470 m off the vessel's observed track. Deliberately not here: it wants the heightmap and the distance-to-shore field, and it has to be done per site because `competition.lock` pins three.
- **#122** — `truth_probe` re-fetches the site centre and convergence live rather than reading this record, with no check on what it gets. Two sources for the same three numbers is what #112 exists to end; out of scope here because the file is #121's.
- **The yaw reference** is the one thing still unmeasured, and `projection.py` says so with the procedure that would settle it.


Ollienel777 (commented):
Mergeable, but fix finding 1 first. CI is green (gate SUCCESS), the branch is CLEAN/MERGEABLE and only one commit behind main, tests/test_site.py passes locally (23 passed), and the PR's central claim — no behaviour change — holds: nothing in whiteout/ imports whiteout.site, and no number moved. The convergence write-up in projection.py is sound; I checked the arithmetic independently (λ−λ₀ = −49.822°, 2·sin(γ/2) = 0.842, and the 0.106° tangent-plane figure is right because it's taken about ARENA_ORIGIN at 71.99/−94.84 rather than the site centre — that's a careful number, not a sloppy one).

Finding 1 is the one that matters: it is latent today, and it fires on exactly the procedure this PR tells the next person to run, at the venue, under a clock.

Findings

Critical

whiteout/site.py:255 — parse_site prefers /api/site's centre within a 50 m bound, but tests/test_site.py:106 pins geometry.py's literals with exact ==. Verified: a centre quoted to 4 dp (4.4 m away, comfortably inside the bound the docstring says exists for exactly that) makes SITE_CENTRE_LAT == site.centre_lat_deg False. Running the documented --write and committing turns the gate red, with no stated answer for whether to hand-edit the literals the PR just called "the copy, not the source".

Medium

scripts/arena_site.py:133 — a 404 on /api/site (an endpoint the PR admits has never been called) raises SystemExit and discards the already-successful /api/env fetch, so the run records nothing — defeating one of the two stated reasons for running the script.
whiteout/site.py:376 — _agree builds a GeoPoint, which raises GeoError, not SiteError. Verified with lat=191.99. It escapes the documented :raises SiteError: and slips past --write's except SiteError, so a swapped lat/lon — the mix-up GeoPoint's validation exists to catch — gives a traceback instead of the designed refusal message.
scripts/arena_site.py:73 — base_url counts colons: http://arena.local/api → http://arena.local/api:8090. Use urlsplit and set the port on the netloc.

Nit

whiteout/belief/geometry.py:348 — the new #: block is DEFAULT_STRAIT prose glued onto SITE_CENTRE_LAT's doc comment (pre-existing on main; this adds ten more lines to it). One blank line before line 349 fixes it.
Design challenges

The tolerant readers are the thing the module warns against. _pick(response, "convergence_deg", "convergence", "grid_convergence_deg") and bounds_3413/bounds are invented spellings for an endpoint nobody has called — "imagination rather than evidence", by site.py's own standard. Since the raw body is kept in the fixture either way, one spelling plus a loud failure naming the keys it saw would cost nothing and would not pre-commit the parser to guesses. Half of _maybe's alias lists are already dead: it case-folds, so SITE_LAT/site_lat are the same lookup.

whiteout/site.py ships in the wheel; its default data source does not. pyproject.toml excludes fixtures*, so load_site() with no argument always raises in an installed wheel. The docstring acknowledges this, but the module then has no product consumer at all — it exists to be read by one test and one script, both of which run from the source tree. tests/ or scripts/ would be a more honest home, and would remove the wheel's only module that cannot work as documented.

50 m of permitted disagreement is a policy, not a rounding. The CENTRE_AGREEMENT_M bound and the exact-equality pinning test encode contradictory beliefs about how much the two centres may differ. Picking one — say, agree to 1 m and compare the constants with pytest.approx at the same scale — would resolve finding 1 and make the intent legible.

Ollienel777:
Review answered, and the fetched record merged in. Gate green, 707 passed.

## The critical finding

`parse_site` now takes the centre from `/api/env` only. `/api/site`'s is corroboration: measured, reported as `centre_agreement_m`, never substituted. The two bounds were answering different questions and one was feeding the other — `CENTRE_AGREEMENT_M` is "are these endpoints describing the same site?", loose at 50 m because four decimal places of latitude is 11 m; the constants check is "is `geometry.py`'s copy still the record's?", exact, because a transcription is either right or it is not. `test_a_coarsely_quoted_site_centre_does_not_move_the_parameters` is the regression.

The other half of that finding was that a real move had no stated answer. It does now in both places it is needed: the test says the record won, edit the literals, re-measure what is derived from them; and `report_drift` prints the same thing on stdout at the moment of the fetch, naming #102's 251×20 → 63×16 and the entropy figures that went with it.

The fetched record happens to agree to 0.0 m, which is luck rather than a reason to have left it.

## The rest

- **404 on one endpoint** no longer discards the other. The failed half keeps the `null` response that already means "never fetched", with the error in its provenance; the good half is written; the run exits 1 saying so.
- **`_agree` → `_place` + `_agreement_m`.** `GeoPoint`'s `GeoError` is spelled as a `SiteError`, so a swapped pair reaches `--write`'s handler as the designed refusal instead of a traceback. Both centres go through it, so a swapped `/api/env` pair is caught even with `/api/site` unrecorded.
- **`base_url` splits with `urlsplit`.** `http://arena.local/api` was becoming `http://arena.local/api:8090`.
- **The nit**: blank line restored between the `DEFAULT_STRAIT` prose and `SITE_CENTRE_LAT`'s doc comment.

## The two design challenges

**Where the module lives** — resolved the other way round: the record moved into the package rather than the module moving out. `whiteout/site.py` shipping with a data source that isn't there was the defect; `[tool.setuptools.package-data]` fixes it at the root, and `load_site()` now works from an installed wheel (verified by unpacking the built wheel outside the source tree and calling it). Moving the module to `scripts/` would have left `geometry.py` citing a script as its source.

**The tolerant readers** — the invented key names are gone (`bounds_3413`, `grid_convergence_deg`, `site_lat`; the last was dead anyway, the lookup folds case). The *shapes* stay tolerated, and 86ce3c7 is why: two of the three this module was written around were wrong, so a list of four is still read beside the object the arena sends.

**One deliberate divergence from 86ce3c7**, flagged so it can be overruled in a line: that commit's comment kept the alternative names because "`competition.lock` pins three sites". Sites are data the same arena software serves, so the key names belong to the software; a second site answering differently is evidence, which is exactly what it takes to add one back. Say the word and they go back in.

## Checking 86ce3c7's arithmetic

Verified against an independent EPSG:3413 implementation. The central claims hold exactly: reprojecting `SITE_LAT`/`SITE_LON` by the closed form lands on the centre of the recorded box to **0.0 m**, the box is square to the last decimal, and the recorded convergence sits **0.0176°** from the identity — 1.00 m across the half-extent, a pass, EPSG:3413 confirmed. That is the premise `projection.py` rests on, now measured rather than assumed.

Two numbers were off and are corrected:

- The span is **0.483%** under the ground extent, not 0.4952%.
- It is **not** the scale factor at the centre, which is 0.580% (0.994202 against the box's 0.995173).

And one thing worth knowing: `bounds3413` is not the footprint of a 6500 m *ground* square. Projecting that square's corners gives an axis-aligned extent of about **9111 m**, because a ground square is turned ~49.8° in this grid and its bounding box is the diagonal — the convergence sighted a third way. So the box is a grid-aligned square whose side the arena derives from `SITE_EXTENT` some way one sample does not settle. It is 31 m across the site, nothing consumes the box, and `BOUNDS_EXTENT_TOLERANCE` holds either reading, so the docstring records it as an open question rather than asserting a cause.

The yaw-against-course-over-ground falsification is still the one thing outstanding, and `projection.py` says so. It wants an arena with the sim running.


---

Ollienel777:
Reconciled against #121, `origin/main` merged in. Gate green, 773 passed.

You were right that this was the blocker, and it turned out to be more than prose: the falsified premise was load-bearing, so the conclusion had to change shape rather than just its wording.

## The critical one

The section asserted "the simulator's world frame is an East–North–Up frame about a geodetic origin, so that yaw is referenced to true North". `truth_probe` measured the opposite. That sentence is gone.

What survives is the frame argument, which was always the load-bearing half and which #121 does not touch: no easting, northing or scale factor of EPSG:3413 is read or written anywhere under `whiteout/`, so no quantity this package computes is expressed in that grid.

What that leaves is one route in, and the section now names it instead of closing it: **whether `ATTITUDE.yaw` is true-north or grid-north referenced is not settled.** The case for true north is now stated as the inference it is — ArduPilot's EKF fuses GPS-derived position against attitude-rotated inertial data, and MAVLink lat/lon is geodetically correct because that is the reference #121 fitted Gazebo's coordinates to, so a grid-referenced attitude would sit 49.8° from its own GPS inside the filter and guided flight would not track; #118 flew the fleet to waypoints. That is an inference from "the vehicles fly", not a measurement, and it says so.

So: no correction is applied because no measurement supports applying one, not because the question is closed. Rotating on the strength of #121's *position* finding would be the same mistake in the other direction and the larger one — a wrong correction costs the same ~1500 m as a missing one.

**One thing #121 gives back:** the check no longer needs a moving asset, which is exactly what defeated the last attempt. Gazebo's `~/pose/info` carries each model's orientation beside the position `truth_probe` keeps, and the two hypotheses differ by exactly the convergence. A parked quadcopter answers it. The conventions need pinning before trusting a sign, so the robust reading is which of 0° or 49.8° the difference lands on.

## The two scope claims

- **"corrects nothing... anywhere else in the build"** → scoped to `whiteout/`. The probe applying the rotation is the distinction rather than a contradiction: it holds Gazebo's grid coordinates, and anything holding a grid quantity must rotate. `bearing_deg`'s note says the same, and now adds the one caller the open question reaches — `arena.py` turns a bearing into a tower's pan, so it is only as good as the tower's own yaw reference. If yaw turned out to be grid north, `bearing_deg` would still be right and that caller would need the rotation.
- **"the only thing that talks to `:8090`"** → false, fixed in `arena_site.py` and `site.py`, and the PR body is updated.

## The mediums

- **truth_probe's duplicate source** → filed as #122 rather than changed here, since the file is #121's. `arena_site.py` points at it.
- **`SECRET_KEY_MARKERS`** → closed, since it is cheap and the repo is public. `is_secret_value` reads the value's shape as well as the key name: issuer prefixes (`pk.`, `ghp_`, `AKIA`, `glpat-`, `-----BEGIN`, …), a URL with inline credentials, and an opaque mixed-case-and-digits token of 32+ characters. Either test firing is enough to redact, so `GITHUB_PAT`, `SENTRY_DSN` and a `DATABASE_URL` with a password are all caught now.

  The shape test is the half that could misfire, so it is tested for quiet as well as for catching: run over every field of the committed `/api/env` record it fires on none of them — the long values there are a dotted registry host, a comma-separated month list and a space-separated colour, and a dot, comma or space fails the character test. A redaction is also a fixed point, so re-running the script over its own output is stable.

## On your local failures

The `imagery.py:267` typecheck and the trigonometry-guard failure don't reproduce here — gate is green on the merged head, 8/8. Agreed they're not this PR's.


---

Issue #44, opened 2026-09-19
ui: every state, exercised


Ollienel777:
**Milestone:** M4 · **Demo beat:** 1 · **Size:** S

## What
Force and screenshot all seven states: no log, loading, malformed log, running,
finished, zero contacts, many contacts.

## Acceptance
- [ ] Each state has a screenshot in the PR
- [ ] The malformed-log state names the failing line and the schema version
- [ ] The empty state offers the bundled fixture as the next action

## Notes
`DESIGN.md`. Depends on #18, #43.

Issue #26, opened 2026-09-19
feat: the scorer, four axes, weights as parameters


Ollienel777:
**Milestone:** M1 · **Demo beat:** 2 · **Size:** M

## What
`whiteout/score/`: our best model of **coverage** (fraction of prior
probability mass observed, weighted by detection quality — not area flown),
**collaboration** (completed cueing chains and cross-class handoffs, from the
contact lifecycle), **efficiency** (distance and energy per unit of information
gained), and **tracking accuracy** (time-in-track and position error while
tracked). Pure function of an episode log. Weights loaded from JSON; fixtures
in `fixtures/weights/` including `equal.json`.

## Why
It is the objective function, the tuning signal and the argument in beat 5.
Weights are parameters so that the booth's answer to Q3 costs minutes.

## Acceptance
- [ ] `score(log, weights)` is pure and returns four finite values in [0,1]
      plus the weighted total
- [ ] No weight is hard-coded anywhere in `whiteout/`
- [ ] Each axis has a hand-computed unit case
- [ ] Re-scoring a cached log with new weights requires no re-simulation

## Notes
`SPEC.md` §1, §3, §11 Q3. Depends on #5, #25.

Ollienel777:
loop: re-triage — the four-axis scorer is wrong, and its purpose is mostly gone

Two changes from `hackathon/ARENA.md`:

**The axes are wrong.** There are **seven** judged criteria, not four: search
efficiency, coverage, detection speed, tracking duration, accuracy, autonomy
and collaboration. No weights are published.

**Its purpose is gone.** This scorer existed so the coordinator could optimise
directly against a faithful model of theirs (#29) and be tuned over thousands
of episodes (#31) — both now closed, because there is one judged run on
Sunday, no leaderboard to calibrate against, and their sim will not run at
100x real time.

**Not closed, because a local measure is still useful** for comparing two
policies during development — is this search pattern better than that one.
But it is now a **development instrument, not a target**, and it should be
rebuilt small against the seven named criteria rather than built faithful
against four invented ones.

**Downgrade it and do it late.** Optimising hard against a guess at someone
else's scoring is worse than a legible policy that can be explained in the
five-minute presentation, which is itself judged.

PR #117, opened 2026-09-20
feat(vision): record whether a frame and its pose were one instant

### LINKED ISSUE #115: feat: record whether a frame and the pose it was projected with were synchronised
A camera projection consumes a frame **and** an attitude together. If the two
are 250 ms apart on the fixed-wing at 22 m/s, that is 5.5 m of error injected
into every fix — with every field populated, nothing null, and nothing to look
at. We do not record whether they were synchronised.

Both numbers are already there. `whiteout/types.py:Pose` carries `measured_t`
(the age of the underlying fix, or `None` when the transport does not know) and
optional `pitch`/`roll`; `whiteout/vision/detect.py:VesselDetection` carries the
frame's `t` and the `pose` it was projected with. The skew is computable today
and nobody computes it.

What this asks for:

- A judgement over the states that are actually distinguishable in our data
  model, plus a max-skew parameter.
- Recorded on the detection, carried through the sighting, and into the episode
  log so the viewer and the scorer can see it.
- A detection whose pose was not synchronised to its frame must be
  **degradable**, not silently trusted: decide refuse-versus-flag and say why
  in the docstring.

Hold the line issues #80 and #101 set: there is no number that would be honest
for something the transport does not report, so absent stays absent rather than
defaulting to zero.

Out of scope: the arena adapter's `measured_t=None` (it has not established the
`time_boot_ms` offset, and that is the correct thing for it to report), and any
sim whose poses are synchronous by construction.

Ollienel777:
Closes #115

## What

`whiteout.types.PoseSync` judges a frame against the pose it is about to be
projected with: a status from `SYNC_STATUSES`, the skew in seconds, and the
`max_skew_s` the verdict was reached against. `PoseSync.for_frame` is the judge.
`VesselDetection` carries one, `Sighting` carries it through, and `Contact`
carries it into the episode log.

`whiteout.types.SightingRefusal` is the other half: a fix the vision path
*declined* to use, with the asset, the tick and the verdict that caused it, on
`EpisodeRecord.refusals`. Without it the two statuses this path refuses on could
never reach a log at all — a camera refused every tick would read as an empty
sea. The rail shows both: a sync cell per contact row, and a
"2 fixes refused — skew 0.41 s, no attitude" note under the list.

Schema 4 → 5 (two required keys: `Contact.sync`, `EpisodeRecord.refusals`); the
viewer's pinned version and the committed demo episode migrated with it.

Four sync states, because four are what this data model can tell apart:
`synchronised`, `telemetry_missing`, `telemetry_stale`, `attitude_missing`.
There is deliberately no `attitude_unsynchronised` — a `Pose` carries one
`measured_t` for the position and the attitude together, so a stale attitude and
a stale fix are the same number here; a separate attitude stamp, if one ever
arrives, gets its own field and its own state rather than a quiet redefinition.

`skew_s` is `None` when the pose reports no measurement time, and also when that
time is not a number: `NaN` fails every comparison and would otherwise be
reported as *perfect synchronisation*. `0.0` there reads as simultaneous, which
is a measurement. #80 and #101 held that line for `measured_t` and for
`pitch`/`roll`.

**Flagged on the detection, refused at the poster, and the refusal is recorded.**
Skew moves the *place*, not the evidence, so a sixth `None` in the detector
would refuse the wrong thing and let one adapter that reports no measurement
time silence a whole camera — the arena's adapter is that adapter, correctly.
So `detect_vessel` only stamps the verdict; `VisionSightings._sight` drops
`telemetry_stale` and `attitude_missing` and *records why*; `telemetry_missing`
is carried and labelled. Argued in `VesselDetection`'s docstring, once.

`frame_t` is in the pose's clock, never the camera's: a live frame is stamped
`time.time()` and a recorded one an index over a frame rate. `sightings` passes
the tick that consumed the frame, which makes the recorded skew a **lower
bound** — the frame's own age inside its feed is still untracked — so
`synchronised` means "no skew this log can see", stated where the verdict is
defined.

Untouched, as the ticket asks: the arena adapter's `measured_t=None`, and
anything whose poses are synchronous by construction.

## Why

Tracking accuracy is one of the four scored axes, and a fix whose frame and
attitude were a quarter-second apart is 5.5 m wrong at the fixed-wing's 22 m/s
while looking exactly like a good one. The log is the only artifact the scorer
and the viewer read, so a position nobody can qualify is a position everyone
trusts — and a fix silently dropped is worse still, because it reads as nothing
to see.

## How to verify

- `python -m whiteout.cli serve`, open the viewer, load
  `fixtures/episodes/demo.jsonl` — validates as schema 5 and the shell reads it.
  The two new surfaces were driven in Chromium at 1440×900 against a log
  carrying all three contact sync states and both refusal reasons (see *Screens
  touched*).
- The four states, precedence, a future-stamped fix, a `NaN`/`±inf` stamp and
  the bound: `tests/test_types_log.py` "frame-to-pose synchronisation".
- Round trip, required-but-nullable keys, an unknown status, a reasonless
  refusal, a refusal stamped from another tick, a non-finite skew: same file.
- Refuse-versus-flag and the refusal record: `tests/test_vision_sightings.py` —
  stale and attitude-less are refused *and recorded*, a frame with no vessel is
  not a refusal, a missing pose or frame is an absence rather than a refusal,
  last tick's refusals do not linger, and a non-positive `max_skew_s` is refused
  at construction.
- Flag-not-refuse in the detector: `tests/test_vision_detect.py` — same pixel,
  same confidence, under every sync state.
- The path to the log: `tests/test_coordinate.py` — a contact carries its last
  fix's `sync`, and what the source refused reaches `TickOutcome`.

## Gate

`python scripts/gate.py`, all eight steps, in this worktree:

```
GATE PASS: install (0.1s)
GATE PASS: lint (0.0s)
GATE PASS: format (0.0s)
GATE PASS: typecheck (0.2s)
699 passed, 12 skipped, 1 deselected in 30.35s
GATE PASS: test (30.6s)
GATE PASS: build (0.8s)
GATE PASS: smoke (0.5s)
GATE PASS: determinism (0.4s)
GATE OK: 8 steps in 32.7s
```

## Screens touched

The rail's contacts panel, checked in Chromium at 1440×900 against a synthetic
schema-5 log (built from the demo episode; not committed):

- each contact row gains a sync cell between the id and the state — `in sync`,
  `no fix time`, `no attitude`, `skew 0.41 s`, or the shell's em dash when the
  log carries no `sync`. Tertiary type, semantic colour only, no new token and
  no raw value. The four-column row fits the 320px rail without ellipsis.
- a note under the list when the tick refused anything:
  `2 fixes refused — skew 0.41 s, no attitude`, amber, shown whether or not
  there are contacts.

No console errors. The bundled demo episode holds no contacts and refuses
nothing, so neither surface appears in it; exercising every state against a real
vision run is still issue #44's job.

## Follow-ups

- `max_skew_s` is one number set for the fastest airframe and applied to towers
  that do not move, where the same skew costs nothing. A per-class bound is the
  honest shape; the docstring says so rather than leaving it to be noticed.
- The frame's own age inside its feed is still untracked, so the skew is a lower
  bound. Closing it needs a frame stamp convertible into the tick clock — the
  same offset problem as #63/#76.
- `Detection` is the natural second home for a `PoseSync` the moment anything
  writes `SensorReport`s. Nothing does today, and a qualifier nobody sets is
  worse than none.
- The scorer does not read `sync` or `refusals` yet; it reads nothing yet (#26).
  A `telemetry_missing` fix is the obvious candidate for a discounted one, and
  refusals are a free coverage signal.


Ollienel777 (commented):
Verdict: mergeable, but I'd fix finding 1 first. Nothing here is a merge-blocker for the demo path (vision isn't wired into cmd_run yet), and the gate is clean on this branch — 685 passed, 12 skipped; lint/format pass; the mypy "unused type: ignore" in whiteout/vision/imagery.py:267 reproduces identically on main, so it's a local Pillow-stubs artifact, not this PR. The demo.jsonl migration is honest: I diffed all 120 records key-by-key and schema_version is the only thing that changed.

Findings
whiteout/types.py:561 — PoseSync.for_frame is documented "pure and total" but raises RecordError on a NaN measured_t: for_frame picks synchronised, then its own __post_init__ consistency check rejects it. Verified on the branch. The call in _sight is outside any handler, so one NaN stamp kills the tick. ±inf yields telemetry_stale with a non-finite skew_s. Both should land on telemetry_missing.
whiteout/vision/sightings.py:263 — _sight refuses both attitude_missing and telemetry_stale, so Contact.sync can only ever be synchronised, telemetry_missing or None. Two of four statuses, both red CSS rules, and the telemetry_stale branch of syncLabel are unreachable.
whiteout/vision/sightings.py:222 — max_skew_s is called the operator's number but is validated nowhere and wired to no CLI flag; a non-positive value fails per-frame at tick time rather than at construction.
On the design

The PR argues its seams well and I agree with flag-on-the-detection / refuse-at-the-poster. But finding 2 is the design challenge worth taking seriously: the refusal deletes the evidence the record exists to preserve. The PR's own framing is "a position nobody can qualify is a position everyone trusts" — yet the two states that say don't trust this are exactly the ones that can never reach the log. Run with --pose-age 0.3 and the vision path goes dark with no contact, no counter, no reason; in the artifact that's identical to an empty sea. That's a worse silence than the one being fixed, because at least an unqualified fix is visible.

The fix is small and doesn't require reopening the seam argument: keep the refusal, but let it leave a trace — a refused-sighting count on the observation, or letting the contact go lost carrying the stale sync. Otherwise the red styling in viewer.css is decoration for a state the demo cannot produce, and the "Screens touched" section is describing a cell nobody will ever see in that colour.

Two smaller notes, not findings: DEFAULT_MAX_SKEW_S = 0.25 is derived from the fixed-wing's 22 m/s, but it's applied uniformly to towers and rovers where the same skew costs nearly nothing — a per-class bound would be more honest than one number wearing an airframe's argument. And the docstrings are doing genuinely good work here, but PoseSync's is now ~60 lines of prose against 40 lines of code; the "no attitude_unsynchronised" and "flagged not refused" paragraphs are each argued twice (here and in VesselDetection), which is the one-rule-one-file convention in CLAUDE.md pulling against itself.

Ollienel777:
All three findings fixed, and the design challenge taken — a14e771.

**1. `for_frame` was neither pure nor total.** Confirmed, and worse than the report: a `NaN` `measured_t` did not merely raise, it first came back `synchronised` — a non-number read as *perfect* synchronisation, which is this record's own failure mode with the sign flipped. Then `__post_init__` turned that into a `RecordError` from inside a tick with no handler over it. `NaN` and `±inf`, in `measured_t` or in `frame_t`, are now `telemetry_missing`, which is what "no usable measurement time" already meant. The docstring now says the totality is load-bearing (one caller, inside a control loop) rather than tidy.

**2. The refusal deleted the evidence.** You are right, and the framing is the one I should have applied to my own seam: an asset that stops contributing for a reason we know is the same silence one level up. I kept the refusal and added the trace, because dropping a known-bad fix into the belief is still wrong — `whiteout.types.SightingRefusal` (asset, tick, verdict) on `EpisodeRecord.refusals`, captured by `VisionSightings` during `sightings()` and read off the source by the coordinator the way `cli` reads a transport's roster, so a one-method source is still a source. The rail now shows `2 fixes refused — skew 0.41 s, no attitude` whether or not there are contacts. Both refused statuses are reachable in a log now, and `--pose-age 0.3` produces a counted, reasoned silence instead of an empty sea.

Two details worth flagging: refusals are checked against the record's clock like every other member (a refusal is built from the newest frame a camera happens to hold, so a stamp from the wrong tick would attribute a camera's silence to the wrong second), and they are *captured* during the call rather than recomputed on demand — a feed keeps only its newest frame, so asking again can pair a different frame with the same pose and report a refusal that never happened.

The red `.contact-row` rules stay, now with a comment saying our own path refuses those two states so in a log this build wrote they arrive on the refusal note instead; the viewer opens logs it did not write, and the data model permits either.

**3. `max_skew_s`.** Validated at construction (`VisionError`), where a non-positive bound names itself, instead of failing inside `PoseSync` once per frame at tick time. No CLI flag, because nothing constructs `VisionSightings` from the CLI yet — when `cmd_run` gains a vision path, the flag goes with it.

**On the two notes.** The per-class bound is right and I did not do it: one number wearing an airframe's argument is now stated as such in the docstring, with per-class named as the follow-up rather than left to be noticed — a tower that cannot move should not lose a fix to a stale pose. And the duplication is trimmed: the flag-versus-refuse argument now lives only in `VesselDetection`, with a pointer from `PoseSync`; the "no `attitude_unsynchronised`" paragraph stays in `PoseSync` alone, since it is about the data model rather than about the detector.

**Verified:** gate clean, eight steps, `699 passed, 12 skipped`. Both UI surfaces driven in Chromium at 1440×900 against a schema-5 log carrying all three contact sync states and both refusal reasons — rows fit the 320px rail, no console errors. The screenshot is not committed; the bundled demo still holds no contacts, so exercising this against a real vision run remains #44's job.


---

Ollienel777:
Fixed in 1b586a9. Reproduced your case first: one refusal bare, zero through `MotionGate(source)` — now 1 and 1.

**The gate forwards.** `MotionGate.refusals()` passes the wrapped source's through, and its docstring now says transparency has to include what the source refused, naming the composition (`MotionGate(VisionSightings(...))`) and what the silence would have cost. You are right that the `getattr` discovery broke the gate's own promise: "the run loop does not know it is there" was true in one direction only.

**The gate's own rejections are still not refusals**, and that is now written down rather than left to be inferred: a candidate dropped for not moving is a judgement about the world, not about our telemetry, and `SightingRefusal` carries a `PoseSync` as its reason, which cannot express "it did not move". `MotionGate.rejected` counts them and nothing writes that to the log. Giving them the same visibility needs a reason type that is not a `PoseSync` — a follow-up, not a widening of this PR. Flagging it because it is the same silence as the one you caught, one layer further in.

**On reconsidering the discovery** — I took the substance of your suggestion but not the mechanism, because neither `@runtime_checkable` nor a defaulting base actually fails loudly: an `isinstance` check against a structural protocol answers exactly what `getattr` answers, and a base class with a default `refusals()` returning `()` would have made `MotionGate` *look* compliant while dropping everything — worse than today. So:

- `refusals` is now part of `SightingSource` itself; the separate `RefusingSightingSource` is gone and `NoSightings` answers both.
- `tests/test_coordinate.py` walks `whiteout/` for every class that reports sightings and fails if it cannot say what it refused. Protocols are skipped, since `motion.Sightable` is deliberately the narrower shape the gate wraps. I verified the guard fires by deleting `MotionGate.refusals` — it fails naming `MotionGate`, which is the signal that was missing when you found this by hand.
- The coordinator keeps its `getattr`, now documented as tolerance for a hand-written double rather than a second contract, with a test that such a double still runs.

Plus the composition test you asked for, at both levels: `tests/test_vision_motion.py` for the gate, and `tests/test_coordinate.py` for source → gate → `TickOutcome.refusals`.

Gate clean: eight steps, `704 passed, 12 skipped`.


---

Issue #13, opened 2026-09-19
feat: the negative-information update


Ollienel777:
**Milestone:** M1 · **Demo beat:** — · **Size:** M

## What
`whiteout/belief/negative.py`: update the field by the likelihood of a
**non-detection**, given range, the camera's field of view, and the asset
class. A camera that swept an area and saw nothing must reduce belief there
**in proportion to how well it could have seen**, never to zero. The docstring
carries the formula.

## Why
The cheapest accuracy win available, and the thing that makes *coverage* and
*search efficiency* — two of the seven judged criteria — quantitative rather
than a heatmap of where we flew.

**It is stronger now than when it was written.** The original specified this
against assumed sensor footprints with a detection probability. The real assets
carry cameras with **published fields of view** — quad 114.6°/99.4°,
fixed-wing 69.0°/42.6°, towers 60.0°/36.1° — and a pose from MAVLink. That
turns "swept here, saw nothing" from an estimate into **geometry**: an exact
cone, intersected with the water, at a known instant.

## Acceptance
- [ ] Property test over ≥50 seeds: belief entropy is non-increasing under
      non-detections alone
- [ ] The posterior stays in [0,1] and normalised, always
- [ ] A sweep at long range reduces belief less than the same sweep close in,
      and the falloff is justified in the docstring rather than tuned to taste
- [ ] A single-camera closed-form case matches a hand-computed value, with the
      hand-working in the test
- [ ] A detector with a known false-negative rate does not drive belief to
      zero on one empty frame

## Notes
`hackathon/ARENA.md` §4. Depends on #12 for the field and **#65** for the
camera geometry that turns a pose and an FOV into a footprint on the water.

**Dropped from the original body:** terrain occlusion (no terrain model of
ours) and the ablation flag for #30, which is closed.

**This remains the highest-severity correctness risk in the build.** An
over-confident non-detection update empties belief from water the vessel is
actually in, and the fleet then never looks there again.

Ollienel777:
loop: re-triage — keep, and it is stronger than when it was written

Negative information was specified against assumed sensor footprints with a
detection probability. The real assets carry cameras with **published fields
of view** — quad 114.6°/99.4°, fixed-wing 69.0°/42.6°, towers 60.0°/36.1° —
and a pose from MAVLink.

That turns "a sensor swept here and saw nothing" from an estimate into
**geometry**: an exact cone, intersected with the water, at a known instant.
[...]

PR #118, opened 2026-09-20
feat(transport): stand up the real fleet, fly it, and rebuild the demo episode

### LINKED ISSUE #43: chore: the committed demo episode fixture

### LINKED ISSUE #71: chore: the kinematic fleet still has a rover, and the wrong callsigns
Filed by the build loop.

**Milestone:** M1 · **Demo beat:** 1 · **Size:** S

## What
`whiteout/transport/kinematic.py`'s stub `FLEET` stands up one asset per
`VEHICLE_CLASSES`, including `rover-1`, and names them `fw-1`, `quad-1`,
`rover-1`, `tower-1`. Bring it to the real fleet:

- **two** towers, one quadcopter, one fixed-wing. No rover.
- callsigns per `hackathon/DESIGN.md`: `WING-1`, `QUAD-1`, `TOWR-1`, `TOWR-2`.
- the per-asset capability data `ARENA.md` §3 and §4 give, as data rather than
  physics: HFOV/VFOV and frame resolution per asset, and that the fixed-wing
  is slightly faster with slightly better sensors while the tower is immobile.

## Why
`hackathon/ARENA.md` is merged ground truth and §3 states the fleet is
**two towers, one quad, one plane, and there are no rovers**. Everything
downstream renders or reasons over whatever the transport names, so the wrong
fleet propagates into the viewer and the policy.

Evidence, from two independent implementers who each hit it and correctly
reported it rather than fixing it out of scope:

- PR #62 (#9, the conformance suite): "`ARENA.md` §3 says the real fleet is
  two towers, one quad, one plane and **no rovers**, while
  `whiteout/transport/kinematic.py`'s `FLEET` stands up one asset per
  `VEHICLE_CLASSES` including `rover-1`."
- PR #70 (#18, the viewer shell): "Callsigns are `fw-1`, `quad-1`, `rover-1`,
  `tower-1` ... The viewer renders whatever the log names — this needs a
  ticket against the transport."

Beat 1 is the first screen a judge sees, and a fleet row reading `rover-1` for
an arena that has no rover is visibly wrong in the one place it is least
affordable.

## Acceptance
- [ ] `FLEET` is two towers, one quad, one fixed-wing, with the DESIGN.md
      callsigns
- [ ] No `rover` remains in the stub fleet or in anything that enumerates it
- [ ] Per-asset FOV and frame resolution are available as data, matching
      `ARENA.md` §4's table
- [ ] The conformance suite (#9) still passes unchanged
- [ ] A test asserts the fleet's composition, so a regression is caught

## Notes
`hackathon/ARENA.md` §3, §4. `hackathon/DESIGN.md` for the callsigns.
Touches `area:transport`, so it queues behind PR #62.
Related: #36 (fleet rows) renders these; #68 (tower control) needs two towers
to exist.

---

Ollienel777:
Closes #71
Closes #43

## What

The kinematic transport published no roster, so `_coordinator_for` returned `None` and **the default `run` produced an empty episode**: four parked assets, no intents, a `1x1` placeholder belief digest with zero mass. The belief field, the policy, the track hold and the tracks client were all merged, tested â€” and unreachable. The seam was finished and nothing was joined to it.

The committed `fixtures/episodes/demo.jsonl` â€” what the viewer opens by default, and the fallback if a live run fails in front of a judge â€” was 120 ticks of exactly that.

## The fleet (#71)

`ARENA.md` Â§3's composition: **two towers, one quadcopter, one fixed-wing, no rover**, with per-asset capability data (camera, cruise speed, mobility).

The `asset_id`s are the arena adapter's `DEFAULT_ROSTER` values, and the test asserts that *equality* rather than a literal list â€” what matters is not which strings they are, it is that an episode recorded on the fake and one recorded on the arena name the same four assets.

**This deviates from #71's stated acceptance**, which asked for the `DESIGN.md` display callsigns (`WING-1`, `TOWR-1`). That ticket predates the arena adapter. A demo whose fleet rows read `WING-1` while the judged run reads `fixed-wing` is two episodes nobody can line up. Callsigns are the viewer's concern.

## The motion

An asset closes on its last commanded waypoint at its class's cruise speed and stops on arrival. A tower is aimed, never sent.

First-order **on purpose**: no turn rate, no bank, no wind, no stall. A fake that reimplements a flight stack badly is worse than one that is obviously a fake, because it invites tuning against its errors. What it models is the only property the policy can observe â€” an asset takes time to get somewhere, and the fixed-wing gets there sooner.

## The fixture (#43)

Regenerated at seed 7 over 400 ticks, with the one command that reproduces it documented in a new `fixtures/episodes/README.md`, and a test asserting what the old fixture would have failed: tasked, moving, and a real belief grid rather than the placeholder.

```
before   every tick    digest {"grid_shape": [1,1], "mass": 0.0}  intents 0  speed 0.0
after    tick 0        covered_fraction 0.12   quad z  40 m (commanded 120 m)
         tick 57       covered_fraction 1.00   quad z 120 m, holding
         tick 57-400   steady state: ~2 re-taskings per tick, entropy unchanged
```

Almost all of the visible change is in the first minute, and the fixture's
README now records that curve rather than leaving it to be found. A 1200-tick
probe confirms the back half stays flat for a reason length cannot fix: see
the known limitations below.

## Two guards caught real mistakes, and both were right

**The no-leak guard** rejected `from whiteout.belief.geometry import DEFAULT_STRAIT`. The obvious way to site a tower is to ask for a point 18% along the strait; the seam forbids a transport importing `whiteout.belief` at all. Where we physically put a mast is a fact about the world, like a port number â€” not the coordinator's model of it. The stations are literal coordinates now, derived offline, provenance in the comment.

**The trigonometry guard** caught the heading's `atan2`. Both endpoints go through `whiteout.geo` first, so it takes an East and a North in metres and returns a bearing: no latitude in, no scale out. It is on `_TRIGONOMETRY_ALLOWED` with that reason. The ellipsoid and conversion-name guards still walk the file with no allowlist, so a second converter here would still have to spell a constant or name itself.

## Known limitations

Two, both recorded in the code rather than left to be found.

**The belief field is uniform for the whole episode and stays uniform.** Nothing erodes it â€” there is no sighting source behind this transport, and the negative-information update (#13) is not merged. A field nobody has looked away from is a fixed point of diffusion. The sweeps are visible; the reason for them is not. That is the gap between a fleet that moves and a fleet that thinks, and it is the highest-value remaining ticket for the demo.

**`whiteout score` still prints four zeroes, and that is not this branch.** `cmd_score` is a hardcoded stub that never reads the log (`scores = {axis: 0.0 for axis in AXES}`) â€” #26. The zeros in the smoke step were two independent problems and only one of them was the transport.

## Gate

CI green. Locally: lint, format and `pytest -q -m "not slow"` pass (666 passed, 12 skipped), smoke runs 400 ticks, determinism byte-identical at seed 7.

Two local-only notes, neither from this branch: `mypy` flags an unused ignore at `whiteout/vision/imagery.py:267` because this machine's venv exposes a system `cv2` (CI has none, where the ignore is load-bearing), and `tests/test_transport_arena.py` needs `pymavlink`, which the gate's `--no-deps` venv omits.


Ollienel777:
Findings 1, 2 and 4 fixed at the latest push; gate green.

**1 — tower ordinal.** Towers are counted among themselves now, and `_tower_station` raises on an unsited index rather than wrapping the list, so a third tower fails loudly instead of standing on top of the first. Two tests: that each mast gets a distinct station, and that the refusal happens.

**2 — commanded altitude.** `_advance` now flies `z`, at a per-class `climb_mps`. You put the distinction better than the PR did, and I took it as written: not modelling altitude is defensible, accepting a `target_z` and contradicting it in the same record is not, because it looks modelled. The quadcopter now climbs 40 m → 120 m over 27 s and holds, and #111's 60 m stand-off is a command the transport can actually obey. Tests cover climb, descent, and arrival without oscillation.

**4 — the speed comment.** Corrected, and it now says explicitly that the policy emits `speed=0.0` on every intent, so this is the branch the whole demo runs on rather than an edge case.

---

**3 — episode length. I checked the hypothesis and it does not hold, so I have not lengthened the run.**

The suspicion is a good one — `stale_horizon_s` is 240 s against a 200 s episode, which really would make decay unreachable by construction. So I ran a 1200-tick probe: 600 s, two and a half times the horizon.

| at 1200 ticks | result |
|---|---|
| `entropy` | `6.745236` at tick 0 and at tick 1199, constant to six decimal places throughout |
| `covered_fraction` | 1.0 from tick 57, one brief dip to 0.96 near tick 1000 |
| re-taskings | ~2 per tick, constant rate, 118 by tick 57 and 2423 by tick 1199 |

The horizon is not the binding constraint. With the belief field flat, the policy's staleness model is the only thing varying, and it saturates at tick 57 regardless of how long the run is — so 800 or 1200 ticks buys nothing but bytes.

**Your underlying point was right, and that part is fixed.** The before/after table did compare tick 0 with tick 400 when the interesting minute ends at tick 57. The fixture README now carries the measured curve — where it flattens, what stays constant, and the 1200-tick probe that rules out the horizon — so the flattening is recorded rather than discovered by whoever opens the viewer next. The before/after table below is rewritten to say where it flattens.

**5 — collision with #117.** Confirmed: it rewrites `fixtures/episodes/demo.jsonl` for schema 5. **Merge #117 first.** Then this branch takes `main`, regenerates with the one documented command, and the conflict never has to be resolved by hand. If this one merges first instead, #117 must regenerate rather than hand-resolve — the file is 400 records of generated output and there is no correct manual merge of it.

---

On `KinematicAsset.camera`: it is unread today, and you are right that a field waiting for a consumer is a hazard. I kept it because it is not speculative — #71's acceptance asks for per-asset FOV and frame resolution as data, and `camera` is the key into `CAMERAS` that supplies both. The consumer is the sighting source. If it is still unread when #13 lands without using it, it should go.

The callsign deviation is now recorded on #71 (https://github.com/Ollienel777/HTN_WHITEOUT/issues/71#issuecomment-5747436984) so the ticket is not read as having met an acceptance it deliberately did not.

PR #123, opened 2026-09-20
feat(belief): a non-detection is evidence, so spend it

### LINKED ISSUE #13: feat: the negative-information update

Ollienel777:
Closes #13

## What

Looking somewhere and seeing nothing is a measurement. The field ignored it, so it was a fixed point of its own diffusion: the committed demo ran 400 ticks with **one** distinct entropy value, and a 1200-tick probe showed the same. The fleet flew sweeps and the sweeps meant nothing.

```
                    before        after
entropy             6.745236  ->  6.7429 -> 5.4582
peak_p              0.001176  ->  0.00122 -> 0.00785  (6.4x)
distinct entropy    1             400
```

## The geometry is exact, not assumed

The ticket was written against assumed footprints with an assumed detection probability. The real assets are better than that: published fields of view, and `ATTITUDE` for all four. So "swept here, saw nothing" is a cone from a known point, intersected with the water.

A cell is looked at when the ray to it lands in frame. Off-frame — including **straight down**, which is #111's finding falling out with no special case — the likelihood is **exactly 1.0**. No evidence, not weak evidence. That distinction is what stops 400 ticks of sweeps elsewhere accumulating into a claim about water nobody saw.

## The falloff is derived, not tuned

Pixels on target go as `h / r³` — one power from angular width, two from the depression angle foreshortening the along-range axis — and detection saturates in the pixel count. **A sweep at twice the range says an eighth as much**, and the reason is geometry plus the statistic `detect.py` actually computes. The test asserts the cube law where the asymptote applies and pins the 7.13 near-field ratio as the saturating term behaving correctly rather than drifting.

## `p_max < 1` is load-bearing, not cautious

`detect.py` has five paths that turn a real vessel into `None`. A non-detection allowed to reach certainty would empty the field of the water the vessel is in, on one empty frame, and `SPEC.md` §4 records that nothing downstream could notice. `SweepParams` refuses `p_max=1.0` at construction. Twenty empty looks are what empty a cell, and the test asserts the decay is geometric in the **odds against unswept water** — renormalisation makes the raw probability understate it.

## #13's first acceptance criterion is false

> Property test over ≥50 seeds: belief entropy is non-increasing under non-detections alone

It isn't, and `test_entropy_can_rise_when_a_sweep_contradicts_the_peak` asserts the counterexample rather than skipping. A sweep over the *peak* that returns nothing moves mass out of it, flattens the field and raises entropy — correctly. We looked at our best guess and were wrong, so we know less than we thought.

The criterion holds **from a uniform prior**, which is asserted over 60 seeds and is the state every episode starts in. That is the claim worth keeping. Asserting the failure rather than skipping it, because a silent skip is how a wrong acceptance survives to be cited by the next ticket.

## Two bugs found on the way

Both fixed here because #13 cannot be correct on top of either.

**`heading` was radians.** The kinematic transport reported 0–2π while the arena adapter converts `ATTITUDE.yaw` out of radians and divides `hdg` by 100, and `CameraPose` reads the field as `yaw_deg`. Every camera was pointed **57× too little** off North. Mine, from #118. There is now a test that asserts a fleet flying east reports 90.

**`pitch` was `None`,** which correctly made `_erode` skip every asset — the first run after wiring it in was still flat. But a fake does not have an *unknown* attitude, it has a *chosen* one: `None` means the transport cannot know, which is the arena's position until `ATTITUDE` arrives, not the fake's. It now flies level and says so. A pose that genuinely lacks attitude is still skipped, and there is a test that it is not assumed level.

Tower altitudes are #108's measured MSL values (116.8 m, 226.5 m) rather than 15 m — a mast on a cliff has a different horizon and a very different footprint.

## Gate

`854 passed, 12 skipped`, lint and format clean, determinism byte-identical at seed 7, fixture regenerated by the documented command.

## Known limitations

**Terrain occlusion is not modelled** and there is no terrain model of ours, so a ridge between a camera and the water reads as swept. This is the one direction the module errs *against* safety; the mitigation is that `p_max` is well below one, so an occluded cell recovers as soon as anything else looks at it. Recorded in the module docstring, not hidden.

**Fog, sea state and ice cover** change the detector's real performance and none is observable from a pose. `scene.py` says light fog compresses contrast by about a third; that lands in `p_max`'s margin rather than in a model.

**`vessel_length_m = 12.0` is the one genuinely assumed number.** It enters squared, so a guess that is too small makes the update *weaker* — the safe direction.


Ollienel777:
All three findings fixed, gate green (`861 passed, 12 skipped`).

## 1 — evidence per second, not per tick

You were right and it was the one that mattered. `diffuse(elapsed)` is per-second; the erosion beside it was per-call, so the equilibrium between them moved with the loop frequency.

A sweep is now credited `elapsed / reference_s` independent looks and the weight is raised to that power. Two frames 0.25 s apart are not two independent measurements of the same water, and the model says so now.

| 100 s of episode, only `tick_seconds` varying | before | after |
|---|---|---|
| `peak_p` at 1.0 / 0.5 / 0.25 s | 0.0033 / 0.0044 / 0.0058 | 0.00328 / 0.00324 / 0.00290 |
| spread | **76%** | 12%, and no longer monotone |

The residual 12% is the policy genuinely re-tasking at a different rate, so the fleet flies a different path — a real difference between those episodes, not an artefact. Held against a **fixed pose** the update is now exactly invariant to how the exposure is split, and that is the test: ten looks of 2 s and forty of 0.5 s give the same field to `rtol=1e-12`.

I added `max_looks` alongside it, which you did not ask for but the same reasoning demands: a stalled loop resuming after a minute is holding a pose from *after* the gap, so crediting sixty seconds of staring at the current footprint would claim evidence about water the asset had already flown past.

**The demo numbers move, and the old ones were the inflated ones**: entropy 6.7429 -> 5.9286 and peak_p 3.5x, against 5.4582 and 6.4x before. Spread across the run — 47.7% of the drop by tick 100, 78.4% by tick 200. The fixture README now says explicitly that these are properties of the model rather than of the tick rate, and records what they were when they were not.

## 2 — `max_range_m`

Implemented rather than deleted. A documented safety bound that does not exist is worse than none: it is a property a reader relies on. And thank you for the catch on the units — the comparison is now against **ground** range, which is what `max_flat_plane_range_m` returns, with a comment saying why, because against slant range it would have rejected a band of cells the bound does not name, by an amount growing with altitude.

Two tests: the default bound is applied at the flat-plane limit, and an explicit one overrides it. Still inert on this strait, which is precisely why it could sit unimplemented without failing anything.

## 3 — the FLEET altitude comment

Rewritten. The `ARENA.md` §3 sentence about the quadcopter starting over the highest point is true of the arena spawn and was the wrong thing to say beside a table whose figures are all heights above the water plane — which is what `_erode` reads every `pose.z` as. It now says that, and says #108 measured MSL and height-above-water to coincide on the arena to a decimetre.

## One process note

The first push failed CI on a `mypy` `no-any-return` in the new code. I had been running `pytest` directly rather than the gate, because this machine's venv exposes a system `cv2` and `mypy` flags an unused ignore in `whiteout/vision/imagery.py:267` that CI never sees. Skipping a gate step because one unrelated line is noisy is how the next real error gets through. Recorded in the fix commit.

Ollienel777 (commented):
lgtm!

Issue #20, opened 2026-09-19
feat: FieldCanvas, the belief field render


Ollienel777:
**Milestone:** M1 · **Demo beat:** — · **Size:** M

## What
One `<canvas>`, layered in a single pass: **the channel outline → the belief
field over water → camera footprints → the four assets → the current track**.

**The erosion animation is the centrepiece**: non-detections visibly carving
luminance out of the belief field as the cameras sweep. It is the one thing
that makes #13 legible to someone who has not read the code.

## Why
This is how the fleet's reasoning becomes visible, and it is the only way
anyone debugs #12 and #13.

It also carries weight in the **five-minute presentation** (#69), which is
judged on how we thought about the problem and on **autonomy** — a judge
watching belief erode where the cameras have looked can see the fleet
reasoning, rather than being told it does.

## Acceptance
- [ ] Renders at the episode's real-time rate with **four** assets, measured
      not assumed
- [ ] Belief is drawn **only over water**; nothing implies the vessel could be
      on land
- [ ] Each asset's camera footprint is drawn from its real pose and FOV, so
      what the fleet has actually seen is visible
- [ ] The current track is distinguishable from belief, with its **age**
      shown — a track held for two minutes and one from four seconds ago must
      not look the same
- [ ] Belief-ramp colours appear nowhere but the belief field
- [ ] `prefers-reduced-motion` snaps between ticks with no animation
- [ ] Nothing glows: no halo shadows, no blur filters, no bloom

## Notes
`DESIGN.md` "The signature element". Depends on #18, #12, #13.

**Rewritten after the API workshop. The original body specified three things
that no longer exist**, and building it as written would have built all three:

- **terrain hillshade** — Dominion Dynamics render Bellot Strait themselves,
  with Mapbox terrain. We do not draw the ground.
- **cut chords** — #14 is closed; a min-cut over a single water channel is
  degenerate.
- **a truth marker** — the real arena gives us **no ground truth**. We never
  know where the vessel actually is, only where we believe it is. Drawing a
  truth marker would be drawing a number we do not have, and the old
  acceptance criterion about `--truth` red went with it.

Also corrected: six assets to **four**, and 20× replay to **real time** —
there is no fast replay because there is no fast simulator.

PR #125, opened 2026-09-20
feat(infra): carry the belief grid in the episode log, so the viewer can draw it

### LINKED ISSUE #124: feat: carry the belief grid in the episode log, so the viewer can draw it
Filed while planning #20. **This blocks #20 and nothing has flagged it.**

## What

`BeliefDigest` carries `grid_shape`, `mass`, `entropy`, `peak_lat/lon/p` and `covered_fraction` — summary statistics only. There is no per-cell probability anywhere in the episode record.

The viewer reads the log and nothing else. So #20's centrepiece — *"non-detections visibly carving luminance out of the belief field as the cameras sweep"* — cannot be drawn. There is nothing to draw.

Anyone starting #20 will discover this about an hour in, after building the canvas layers that *can* be drawn.

## Acceptance

- [ ] The episode record carries the belief field's water cells, per tick
- [ ] Quantised, not raw floats. 850 water cells as `uint8` + base64 is **1133 B/tick, 443 KB for a 400-tick episode**; the same as JSON floats is 6.5 MB, and #43 caps every committed episode at 25 MB combined
- [ ] The cell -> lat/lon mapping is recoverable by the viewer without reimplementing `ChannelBeliefGrid` — either the centres ship once in a header record or the geometry does
- [ ] Round-trips: written, read back, and the reconstructed field is within one quantisation step of the original
- [ ] `whiteout/log.py`'s schema version goes up, and the viewer's `SCHEMA_VERSION` with it
- [ ] The committed fixture is regenerated (`fixtures/episodes/README.md` has the command)

## Notes

Quantisation is lossy and that is fine — this is a **rendering** channel, not a scoring one. Nothing reads it back into the belief field. Say so in the docstring so nobody later mistakes it for state.

Worth considering: ship the field every N ticks rather than every tick. The viewer interpolates or holds. At 443 KB it is not yet worth the complexity, but at 1200 ticks it would be.

Depends on: #12 (the grid), #13 (which is what makes the field worth drawing, PR #123).
Blocks: #20.

chessdudeo7:
## What

`BeliefDigest` carried summary statistics only, so the belief field's per-cell probability never left the process. The viewer reads the episode log and nothing else, which meant #20's centrepiece — non-detections carving luminance out of the field — had nothing behind it to draw.

Every record now carries a `BeliefFrame`.

## The encoding, and why each part of it

**One byte per water cell, base64.** 850 cells is 850 B, 1136 base64 characters, about 1.2 kB a tick with the JSON around it and **504 kB over a 400-tick episode**. The same 850 cells as JSON floats is 6.8 MB. #43 caps every committed episode at 25 MB combined.

**Re-scaled per frame, not over a fixed `[0, 1]`.** A field over 850 cells peaks near 0.01 even when it is sharply concentrated, so a fixed range would spend 254 of its 255 codes on values that never occur and quantise the whole field to 0 or 1. `scale` is the peak, so a byte `b` decodes to `b / 255 * scale` and the round trip is within one step. There is a test that fails if someone replaces `scale` with a constant.

**The cell → lat/lon mapping ships as a water mask plus a cell-corner lattice, once, on the first frame.** Corners and not centres because the consumer draws quadrilaterals: a centre needs the cell's local bearing to be drawn as anything but a dot, and that bearing is a trigonometric question about the channel — the one thing the viewer may not answer for itself, since `whiteout/geo.py` is the repository's only converter and the JS cannot import it. Handing over corners hands over the drawing without handing over a second conversion.

**Big-endian `float32` on the wire**, so the bytes do not depend on the machine that wrote them. `float32` resolves about 1e-5 degrees here, against cells 100 m across.

**The layout is episode-constant, so it is written once.** It is 11.8 kB; restating it on all 400 ticks would cost 4.7 MB to say the same thing 400 times. That is the one place this log departs from "every line is self-describing", and `water_cells` is carried on every frame so that a frame still validates its own length.

`iter_episode_log` now enforces the other half of that trade: the first frame carrying a field must carry the lattice. Without it, `tail -n +2` over a committed episode yields a log where every line validates, every frame decodes to the right length, and nothing can be drawn.

## Acceptance

- [x] The episode record carries the belief field's water cells, per tick
- [x] Quantised, not raw floats — measured at 504 kB against 6.8 MB for the episode
- [x] The cell → lat/lon mapping is recoverable by the viewer without reimplementing `ChannelBeliefGrid` — the corner lattice ships once
- [x] Round-trips within one quantisation step, asserted on a field with structure in it rather than on the uniform prior, and again through JSON
- [x] `whiteout/log.py`'s schema version goes to 6, and the viewer's with it
- [x] The committed fixture is regenerated — 730 kB → 1.24 MB

## Notes for the reviewer

**This is a rendering channel and not a state channel**, said in the docstrings and enforced: a test asserts `whiteout/belief/encode.py` never builds a field back out of one. The decoders exist for the tests and for reading a committed log by hand. Checked directly as well as by test during review: `belief_frame()` leaves every array on the grid bit-identical and `entropy`/`mass`/`peak` unchanged.

**The round-trip tests deliberately do not run on the uniform prior.** A flat field round-trips through any encoding that preserves a constant, so every round-trip test here folds in a detection first, and a separate test asserts the encoded codes actually span the range. The uniform case is asserted as the thing it is.

**`demo.jsonl` now varies every tick.** This branch was opened before #123, and an earlier version of this section said the opposite — 850 identical bytes a tick, because nothing eroded the field until #13 landed. #123 landed it, main is merged in, and the fixture is regenerated against the merged tree: **400 of 400 cell payloads distinct**, entropy 6.7429 → 5.9286, peak probability up 3.5×. `fixtures/episodes/README.md` carries the numbers and keeps the old claim as the reason that section is measured rather than remembered.

Two environment notes, neither of them in the diff: the gate needs Python 3.11 (`pyproject.toml` pins `<3.12`) and `node` on PATH for `test_the_viewers_projection_returns_the_same_metres_as_whiteout_geo`.

Closes #124


Ollienel777:
Picked this up. `main` merged, all three conflicts resolved, both criticals and the medium fixed. Gate green: 8/8, 937 passed, 12 skipped. `mergeable_state` is `clean`.

Worked in the order the review prescribed.

## The conflicts

**`whiteout/coordinate.py`** — both sides added an import and both are used; combined.

**`fixtures/episodes/demo.jsonl`** — regenerated on the merged tree with the README's own command, not conflict-resolved. Measured after: **400 of 400 cell payloads distinct**, 400 distinct entropy values, 6.7429 → 5.9286. The diagnosis was exactly right, and so was the warning about how it lands: the determinism step compares two fresh runs to each other rather than to the committed bytes, and nothing asserted the demo's belief varies, so picking a side would have shipped a flat field behind a green gate.

**`fixtures/episodes/README.md`** — neither side was right alone. The "What is actually missing — Nothing erodes the belief field" section is false as of #123 and is gone; its 1200-tick probe paragraph was already carried by main's text. What this branch contributes is the other half, now measured: the field is in the log *and* its bytes change every tick. The old claim is kept as the reason the section is measured rather than remembered.

Every figure in that file was re-checked against the regenerated episode rather than carried across — the entropy pair, the peak ratio, both erosion percentages, 850 water cells, the 11.8 kB lattice, the sizes. The size line moves: **1.24 MB, of which 504 kB is the belief field** (492 kB of cells plus the lattice). The PR body's 480 kB is updated to match.

## The medium — `log.py`

`iter_episode_log` now refuses a log whose first field-carrying frame has no geometry, beside the `t`-monotonicity rule, since `validate_line` sees one line at a time and cannot know it is looking at the first. Your `tail -n +2` case is the test, run against the committed episode rather than a synthetic record:

```
line 1: the first belief_field in the log carries no geometry, so its cells have no
positions and the field cannot be drawn. The lattice is written once, on the episode's
first frame, so this is what a log truncated at the front looks like
```

The positive control asserts the other half — exactly one record carries the lattice — and a third test pins that a log with no belief field at all is still valid, which is what the kinematic transport wrote before #124.

## Two things I checked that you flagged as clean, and one you didn't

**The rendering-channel claim holds under direct test, not just by the existing assertion.** I snapshotted every ndarray on a worked grid, called `belief_frame` three times, and compared: all twelve arrays bit-identical, `entropy`/`mass`/`peak` unchanged.

I went looking because of something that turned up on the way: **the regenerated episode differs from main's in the last bit of `entropy`, from record 16 on, 1.8e-15 at worst.** That is not this branch. Main's own committed fixture does not reproduce on this machine either — I ran main's code against main's fixture and it diverges at the same record. It is a numpy build difference between whoever generated the previous commit's file and this environment.

It does make one sentence in the README false in both versions, so it is corrected: "regenerating at seed 7 gives a byte-identical file" holds on one machine and not across them. The gate never compares against the committed bytes, so it is not a determinism failure — but the next person to `cmp` this file deserves to know that before they go hunting.

**Your `imagery.py:267` typecheck failure does not reproduce here** — 36 source files, no issues. Agreed it is not this PR's.

## Untouched

The two scope notes — the viewer not drawing the field yet, and per-frame rescaling showing relative contrast — are left as they are. Both are consistent with the title, `scale` is carried so an absolute mode stays available, and neither is a defect to fix in this PR.


---

Ollienel777 (commented):
lgtm!

Issue #30, opened 2026-09-19
feat: the baseline-versus-policy comparison harness


Ollienel777:
**Milestone:** M1 · **Demo beat:** 2 · **Size:** M

## What
`whiteout/tune/compare.py` and `whiteout.cli ablate`: run N seeds against a set
of policy configurations, score each with #26, and emit a table with means and
confidence intervals. Ships with the ablation set: full, negative-information
off, choke-points off, hysteresis off, info-gain off, and frontier baseline.

## Why
This is how we find out whether information-gain routing actually beats a
lawnmower sweep under the booth's weights, instead of guessing. It is the
argument in beat 5 and the evidence in the README.

## Acceptance
- [ ] `whiteout.cli ablate --seeds 50` emits a table to `artifacts/`
- [ ] Every ablation flag from #13, #14, #16, #17 is wired
- [ ] Results are reproducible from the seed list alone
- [ ] The table states plainly when a differentiator does **not** help

## Notes
`SPEC.md` §10 (first risk). Depends on #26, #13, #14, #27, #16, #17.

Ollienel777:
loop: re-triage against `hackathon/ARENA.md` — the baseline-versus-policy harness needed a fast local sim to run both over many episodes. See #25 and #29. A simple A/B against their sim is still worth doing if time allows, but not as specified.

PR #126, opened 2026-09-20
docs(submit): the README, the demo run-through, and SPEC's lost re-plan

### LINKED ISSUE #42: submission: the demo video
**Milestone:** M5 · **Demo beat:** 1–7 · **Size:** S

## What
A three-minute capture of the §2 beats at 1280×720.

## Why
Optional but recommended, and cheap once the beats work.

## Acceptance
- [ ] Under three minutes, 1280×720, audible
- [ ] Beat 3 is legible at video bitrate

## Notes
**Fifth on the cut list** (`SPEC.md` §9 item 5). Depends on #41.

### LINKED ISSUE #49: docs: the README and the four-axis argument
**Milestone:** M5 · **Demo beat:** — · **Size:** M

## What
The README a sponsor engineer opens: the architecture diagram, the transport
seam, the four differentiators with their formulas linked, the ablation table
from #30, and a plain statement of method — the four axes are the sponsor's own
published criteria, the weights are parameters, and the table shows which
mechanism earned which points.

## Why
Optimising a published objective is engineering, not gaming, and saying so
plainly is cheaper than being asked.

## Acceptance
- [ ] Architecture diagram matches `SPEC.md` §4
- [ ] The ablation table is real output, not illustrative
- [ ] Every claim in it is checkable against a committed artifact
- [ ] No attribution stamps anywhere

## Notes
`SPEC.md` §10 (last risk). Depends on #30, #48.

Ollienel777:
Closes #49
Closes #42

Written while #124 is in flight elsewhere. No overlap: this touches `README.md`, `hackathon/DEMO.md` and `hackathon/SPEC.md` only.

## 1. `README.md` (#49)

It did not exist. It is the source link in the submission and the first thing a sponsor engineer opens.

It leads with **the idea**, not the stack — a corridor is not an area, and a non-detection is a measurement — then the seam diagram, then four ideas worth reading the code for, and only then the file tree.

Two things it does deliberately:

**Every number carries the command that checks it.** #49 asks that every claim be checkable against a committed artifact, so there is a table of claim-to-command rather than prose assertions.

**A "what is honest about the state of this" section**, naming the stub scorer, the uncompressed-imagery gap in the detector, unmodelled terrain occlusion, and the absence of ground truth. A reader finds those anyway, and finding them unannounced is worse than being told.

## 2. `hackathon/DEMO.md` (#42)

`PRESENTATION.md` is what you *say*. This is what you *do*, keystroke by keystroke, plus what to do when a step fails.

The preflight's second check is `not 1 distinct entropy value` — **which is exactly the state `main` is in right now.** The check earns its place by failing today. Verified:

```
--- preflight 2 ---
1 distinct entropy values
```

It also says plainly that **beat 2 cannot be performed until #124 and #20 land**, and that rehearsing a beat 2 which narrates a static screen reads worse than admitting the gap. Better the other session knows why their ticket is the demo.

Every command in it was run against this checkout: `run`, the entropy probe, `serve` (prints its URL), the `no_leak` test selector, and `WHITEOUT_TRANSPORT` as the actual env var name.

## 3. `SPEC.md` §4 and §10 — recovered

**PR #75's squash landed only the §1 commit.** `main` still described `whiteout/sim/`, `whiteout/tune/`, a `Transport (ABC)`, `sitl` as validation and `arena` as the speculative third implementation — all superseded by the API workshop, and none of it true since.

That blocked this PR directly: #49's acceptance asks the README's diagram to match §4, and §4 was wrong. Cherry-picked `dc22b6e`; it applied clean. The recovered §4 carries the inverted seam, the `Track` second path, the `vision/` and `tracks/` packages, and the two corrections — *"there are no external services at runtime" was false*, and **there is no ground truth**. §10's risk table comes back with it, superseded table kept below the live one.

Worth a look at how that was lost: the branch had two commits, the squash message named only the first, and nothing checks that a merged PR's content matches its final diff.

## Acceptance not met

**#49 asks for the ablation table from #30.** That issue is closed and the baseline-versus-policy comparison harness with it, so the bullet cannot be met. The claim-to-command table is what stands in its place — weaker, and honest about being so.

**#42's acceptance is about a produced video**, not a script. This closes the part that can be written down; the capture itself is a human action and the document is what it is captured from.

## Gate

`773 passed, 12 skipped`, lint and format clean. Docs-only otherwise.


Ollienel777 (commented):
Findings
hackathon/SPEC.md:217 — high. "whiteout/score/ survives only as a development instrument for comparing two policies, rebuilt small against the seven real criteria" is wrong four ways: the package doesn't exist (scoring is cmd_score in cli.py), cli.py:228 sets scores = {axis: 0.0 for axis in AXES} so it returns zeros for every log and can compare nothing, and AXES is still the original four axes, not seven. This PR's own README.md:174 says the opposite and is correct — "The scorer is a stub… prints four zeroes" — as does SPEC's own risk table at line 555 ("that arbiter is gone"). Three contradictions, one of them inside the same file.
hackathon/DEMO.md:43 — medium. "[#124](https://github.com/Ollienel777/HTN_WHITEOUT/issues/124) — the episode log must carry the belief field. Today it carries summary statistics only" is false: [#124](https://github.com/Ollienel777/HTN_WHITEOUT/issues/124) merged as #125, and every record now carries 850 quantised cells with geometry on the first. The conclusion (leave a hole where beat 2 goes) still holds, but only because the other blocker is the real one — drawField() at viewer.js:667 never reads belief_field, so #20 genuinely hasn't landed. The presenter is told the wrong half is missing.
hackathon/SPEC.md:179 — nit. │ that counts**│ is 59 chars where the rest of the box is 58, so the border steps out a column; the ** also renders literally inside a fence. README's copy of the same diagram uses plain text there and measures 58 throughout.

Ollienel777:
All three fixed, plus one more of the same class. `main` merged in; gate green, `937 passed, 12 skipped`.

## 1 — the scorer claim (high)

You were right on all four counts and I checked each: no `whiteout/score/`, `cli.py:232` is `scores = {axis: 0.0 for axis in AXES}`, `AXES` is a `tuple[str, str, str, str]`, and the README forty lines away says the opposite.

That paragraph came in with the cherry-picked `dc22b6e`, written at the workshop before anyone had opened `cmd_score`. It was wrong when it was written and I carried it forward without checking — which is the part worth noting, since the whole point of recovering that commit was that `main` had a stale §4.

Replaced with what is true, and with **why it is not a gap to be filled**: the sponsor scores what we POST, from data we have no ground truth for, so a scorer of our own would be a number we invented grading a run we cannot check. That follows from §10's own risk table rather than contradicting it.

I also took the `replay` point. The CLI row now marks both stubs rather than keeping one and dropping two — `replay` has always printed *not implemented yet*.

## 2 — the wrong blocker (medium)

Corrected, and thank you: telling a presenter the wrong half is missing is worse than telling them nothing. Verified both halves of your finding —

```
$ python -c "...json...[belief_field]"
['cells', 'geometry', 'scale', 't', 'water_cells']
$ grep -n belief_field viz/viewer.js
50:    "contacts", "truth", "refusals", "belief_field"
```

One mention, in a list of record keys, and `drawField()` never reads it. **#20 is the sole blocker**, and the document now says so with the command that shows the data is already there.

## 3 — the diagram (nit)

Fixed, and it had a sibling. While correcting `that counts**` I measured every row and found the **Track line steps from column 50 to 49** for four rows running — in *both* copies of the diagram, so the README's was not the clean one either. Aligned in each, and the two now agree character for character. Verified programmatically rather than by eye:

```
README.md:        track column(s) [50]
hackathon/SPEC.md: track column(s) [50]
```

## One note on the preflight

DEMO.md's check 2 was written to fail on `main` and did: it printed `1 distinct entropy values`. With #123 merged it now prints `400`. The check has stopped firing because the thing it was watching for got fixed, which is the outcome you want from a preflight and not a reason to remove it.

Issue #109, opened 2026-09-20
fix: the quadcopter's gimbal is not commandable, so its camera looks where the airframe does


chessdudeo7:
Tried live on the arena:

```
MAV_CMD_DO_MOUNT_CONFIGURE  -> MAV_RESULT_FAILED
MAV_CMD_DO_MOUNT_CONTROL    -> MAV_RESULT_FAILED
```

The quadcopter carries `gimbal_small_2d` in Gazebo, but the ArduPilot instance
has no mount configured, so **nothing we send points it**. No `MOUNT_STATUS`
or `GIMBAL_DEVICE_ATTITUDE_STATUS` is sent either, so we cannot read where it
is pointed. It is a fixed camera at the airframe's attitude.

## Why it matters

Flown to 150 m over the channel, the quadcopter's camera was looking at the
**horizon**, not the water: shore, cliff and sky. A frame grabbed directly
over the vessel's last position was mostly terrain.

That breaks the role the policy gives it. `whiteout/policy/search.py` makes the
quadcopter the asset that "can stop and stare", and hovering directly over a
contact is the worst place to put a horizon-looking camera — the contact is
straight down, where it cannot see.

## What to do about it

**Stand off, do not overfly.** A forward-looking camera sees a contact at a
slant range, so the hold should place the quadcopter at an offset and an
altitude that puts the target in the frame, not at the target's own lat/lon.
The offset follows from the published VFOV and the altitude, and
`whiteout/vision/projection.py` already has the geometry to compute it.

**The airframe is the only aiming lever there is.** A multirotor pitches nose
down to accelerate, so a moving quadcopter looks further down than a hovering
one. That is crude and not worth modelling, but it explains why frames differ
between transit and hover.

**Check the towers separately.** They are AntennaTrackers, their pan and tilt
are SERVO1/SERVO2, and those are a different mechanism from a mount — `#68`
owns that and this issue does not cover it.

## Acceptance
- [ ] The hold places the quadcopter at a stand-off, not over the contact
- [ ] The stand-off is derived from the camera's VFOV and the altitude, not a
      guessed constant
- [ ] A test pins that a contact at the commanded stand-off projects inside
      the frame

PR #127, opened 2026-09-20
feat(viz): FieldCanvas, the belief field render

### LINKED ISSUE #20: feat: FieldCanvas, the belief field render

chessdudeo7:
**Stacked on #125 (issue #124), which is its base.** #124 is what puts the belief field in the episode log; without it there is nothing to draw. GitHub will retarget this to `main` when #125 merges. The diff shown here is this ticket's work alone.

## What

One `<canvas>`, one pass, five layers in the ticketed order: **the channel outline → the belief field over its water → each asset's camera footprint → the four assets → the current track.**

## The decision worth arguing about

**The ramp is indexed by how much likelier a cell is than chance, not by its raw probability.**

A field over 850 cells that has learned nothing still sums to 1, so every cell holds about 0.0012 and the brightest cell is brightest only by rounding. A ramp over the raw values paints a uniform field at full luminance and says *the vessel is certainly everywhere* — which is how the first version of this looked, and it was a white slab.

Against chance it says the true thing. Unlooked-at water sits at 1× and draws in the ramp's low blue — DESIGN.md's "flat and dim". Water a camera swept and found empty falls below 1× and drains toward the transparent floor. Water a detection landed in climbs. The picture is **information**, which is what the fleet is actually accumulating, and it means the erosion #13 produces will be the visible event rather than a rounding difference in the top eighth of a ramp.

## The footprint

Drawn as the frame's own ground quad: depression θ ± VFOV/2 off the pose's pitch, ranges `h / tan(θ)`, swept through HFOV. The near and far edges are **pinned under `node` against `whiteout.vision.projection` at the same two pixels** — a footprint computed off the boresight instead of the frame edges looks entirely plausible and is wrong by hundreds of metres, and that is the failure this catches.

Bounded by **the arena's extent**, not by the flat plane's 12.4 km model limit, because `project_pixel_to_ground`'s own docstring asks a caller who knows the arena to do exactly that. At the model limit the quadcopter's wedge covered the whole canvas and claimed ground two arenas away.

Two edge cases are tested rather than assumed: a **level** camera (the quadcopter's, per #109) has no far edge to project, so it stops at the bound; and a camera **pitched past its own nadir** — normal on a 99.4° lens — starts at zero rather than at a negative range that would draw the wedge inside out.

## Acceptance

- [x] Renders at the episode's real-time rate with four assets, **measured**: 2.9 ms a frame over all 400 ticks, 850 water cells, 876×308 CSS px at DPR 1.25, three runs agreeing to 0.06 ms. The tick is 500 ms. Recorded in the source beside `drawContent`, and re-measurable through `WHITEOUT_VIEWER.drawField`.
- [x] Belief drawn **only over water** — the quads come from the mask's set bits and from nothing else, and a frame whose length disagrees with the layout draws nothing rather than the wrong water
- [x] Each asset's footprint from its **real pose and FOV**, with the FOV table pinned to `whiteout.vision.camera.CAMERAS`
- [x] The track is distinguishable from belief — a hard accent cross where belief is a soft wash — and carries **the age of its fix**, which the viewer measures from the log since no field records it. Past ten seconds the ring goes and the cross goes dashed.
- [x] Belief-ramp colours appear **nowhere but the belief field**, asserted as "read in exactly one place"
- [x] `prefers-reduced-motion` snaps between ticks — the replay clock already advances by whole ticks with no interpolation, and nothing added here animates
- [x] **Nothing glows** — no `shadowBlur`, no `ctx.filter`, no `globalCompositeOperation`, no CSS blur

## Notes for the reviewer

**`tests/test_geo.py`'s trigonometry allowlist gains `tests/test_viz_shell.py`**, with a reason written beside the two sibling entries. The `radians` calls are on a camera's pitch and a field of view; the one position in the file goes through `whiteout.geo` like every other caller. I took the allowlist rather than open-coding `* pi / 180`, because dodging a guard is worse than being on its list where a reviewer can see it.

**Batched by colour, one `Path2D` per distinct code.** This is not only speed. Filling 850 quads one at a time leaves an antialiased hairline of background along every shared edge, and the field reads as woven fabric — visible in the first screenshot I took. A single fill over a path that already holds both neighbours covers the shared edge exactly once.

**What this does not do.** No intent lines and no truth marker: the ticket's layer list has neither, and the arena gives us no ground truth to draw. The track layer is unexercised by `demo.jsonl`, which has no sighting source behind it and so no contacts — it is covered by the age measurement's unit test rather than by the fixture.

`node` must be on PATH for the footprint pin; it already was for the existing geo pin.

Closes #20


Ollienel777 (commented):
Not mergeable as-is — a units bug in two of the new layers

MERGEABLE / CLEAN and gate SUCCESS, and it merges with main cleanly (946 passed, 12 skipped; lint/format/build/smoke/determinism all green). Mechanically it's ready. But the footprint and asset layers read pose.heading in the wrong units, and nothing in the suite catches it.

Findings
viz/viewer.js:986 — critical. groundFootprint takes pose.heading straight into screenAngle, which is pure radian arithmetic, but Pose.heading is degrees — settled by sightings.py:346, yaw_deg=pose.heading. The committed episode's first record carries 279.25, 314.48, 106.49, 282.26. Read as radians those wrap to 159.67°, 18.33°, 341.43°, 332.31° — off by 240°, 64°, 235° and 50°. Line 1188 repeats it for the asset heading tick. The half-FOV beside it is converted correctly (camera.hfov * Math.PI / 360), so every wedge is the right shape and size and points the wrong way — the version of this bug that is hardest to catch by eye.
viz/viewer.js:991 — critical, latent. Same error on pitch (pitch_deg=pose.pitch in the same place), which sets the near/far ranges. Invisible today because every fixture pose reports pitch 0.0. Against the arena a realistic −12° arrives as −12 radians, wraps to ~84° nose-up, trips lower <= 0, and the wedge silently vanishes. The fixture looks fine right up until beat 4 points this viewer at the real thing.
tests/test_viz_shell.py:505 — medium. The new test that pins the footprint against project_pixel_to_ground passes {"heading": 0.0, "pitch": math.radians(pitch)} — it converts to radians before crossing into JS and fixes heading at the one value where both units agree. So it validates the geometry in metres while baking in the wrong unit contract, which is why the suite is green with two layers drawing wrong.

PR #133, opened 2026-09-20
feat(score): derive the axes from the episode log instead of printing zeros

### LINKED ISSUE #130: feat: whiteout score prints real numbers instead of four zeroes
`cmd_score` is a hardcoded stub (`whiteout/cli.py:217`):

```python
scores = {axis: 0.0 for axis in AXES}
```

It reads the log only to count lines. Every run prints `0.0000` four times and a total of `0.0000`.

## What

This is the single most damaging thing on screen in the demo. A judge who sees four zeroes next to an architecture narration concludes the architecture does not work. Three honest numbers beat a paragraph.

**Scope it to what the log already carries.** Do not wait on #26's full scoring model.

Derivable from an existing `EpisodeRecord` with no new plumbing:

- **coverage** - `belief_digest.covered_fraction` at the last tick
- **tracking duration** - ticks with a non-empty `contacts`, over total ticks, in seconds
- **detection speed** - `t` of the first record carrying a contact, or "not detected"
- **search efficiency** - total `observation.poses[].energy_used` at the last tick, against coverage reached

**Accuracy is blocked and should be left out for now.** `cmd_run` writes `truth=Truth(t=observation.t, targets=())` - always empty - so there is no truth in any log to measure against. `scripts/truth_probe.py` (#121) reads the live arena, not a log. Populating `Truth.targets` during an arena run is a separate ticket and a bigger one; print `accuracy: not measured (no truth in log)` rather than a zero.

## Acceptance

- [ ] `whiteout score LOG` prints a real value for every axis it can compute from the log
- [ ] An axis that cannot be computed says so in words, never `0.0000`
- [ ] Each printed axis names what it was derived from, so the number is falsifiable
- [ ] The existing `--weights` path still works and the weighted total uses only computed axes
- [ ] A test over `fixtures/episodes/demo.jsonl` asserting the axes are non-zero and bounded
- [ ] `AXES` and the seven criteria in `hackathon/ARENA.md` section 5 are reconciled, or the mismatch is stated in the output

## Notes

`ARENA.md` section 5 says seven criteria - search efficiency, coverage, detection speed, tracking duration, accuracy, autonomy, collaboration - and `AXES` carries four. Autonomy and collaboration are read off behaviour and explanation, not off a log, so they are correctly not here; say that in the output rather than silently dropping them.

Related: #26 (the full scorer). This is the demo-sized subset, not that.

Ollienel777:
Closes #130

Opened by the build loop.

## What

`whiteout score` derived nothing: it set `scores = {axis: 0.0 for axis in AXES}` and read the log only to count lines, so every run printed `0.0000` five times over. It now reads the episode log back through `read_episode_log` and derives, in a new `whiteout/score/` (the path `SPEC.md` §5 already assigns it):

- **coverage** — `belief_digest.covered_fraction` at the last tick.
- **detection_speed** — how early the first record carrying a contact appears, as `1 - (t_first - t_start) / span`. No contact anywhere in the log prints `not detected`.
- **tracking_duration** — the share of ticks carrying a contact in a state that means we are holding it (`tracked` or `handed_off`), with the equivalent in seconds in the derivation line. A `lost` contact is not a tracked tick: `TrackHold._last` is never cleared, so `Coordinator` emits one on every tick after the first sighting for the rest of the run, and counting non-empty `contacts` would measure time since first detection instead. A log carrying no contact at all prints `not detected` — the same absence `detection_speed` reports, decided once rather than twice.
- **search_efficiency** — `covered_fraction` averaged over the episode weighted by each tick's increment of `observation.poses[].energy_used`, differenced **per `asset_id`** so an asset that misses a tick does not land its whole cumulative spend on one tick's coverage. Bounded in `[0, 1]`, unlike a bare coverage-per-joule ratio. It is a convex combination, so it is scale-invariant in energy: it ranks *when* the energy went, not how much of it there was, and the module docstring and the derivation line both say so. A log where no energy changes prints `no energy recorded`.
- **accuracy** — not derived. `cmd_run` writes `Truth(t=..., targets=())` on every tick, so no log holds a target to compare against; it prints `not measured (no truth in log)`.

Each axis prints its value on its own line and the field it came from on the next, indented. The weighted total is a weighted mean over the axes the log answered for, so an axis without a number neither lifts nor drags it. The closing line reconciles `AXES` with `ARENA.md` §5: five of the seven criteria are axes, and autonomy and collaboration are named as read off behaviour and the explanation rather than off a log.

Four files follow that axis list rather than the old four: `fixtures/weights/equal.json` (five keys at 0.2), `hackathon/SPEC.md` (§3's criteria table row, §5's scorer row and §6's smoke row), `docs/arena.md`'s arena-day line about what `score` prints, and `README.md`'s honesty section, which said the scorer prints four zeroes without reading the log. `scripts/gate.py` follows it too. The gate's smoke check changed shape — it required five finite floats, which an axis printing words cannot satisfy. It now requires a line for every axis, requires `coverage` to be a finite number, rejects any other axis that is numeric but non-finite, and accepts words for the rest.

This is the demo-sized subset, not #26's scoring model.

## Why

Issue #130: `score` is what a judge sees beside the architecture narration in the live three-minute demo, and four zeroes there read as an architecture that does not work.

## How to verify

```
python -m whiteout.cli score fixtures/episodes/demo.jsonl --weights fixtures/weights/equal.json
```

on this branch, against the committed 400-record episode:

```
coverage: 1.0000
    belief_digest.covered_fraction at the last tick (t=199.5 s, grid 63x16)
detection_speed: not detected
    no record of 400 carries a contact, so there is no first detection to time
tracking_duration: not detected
    no record of 400 carries a contact, so there was never anything to hold
search_efficiency: 0.9833
    coverage averaged over the 7581 units of observation.poses[].energy_used spent after t=0.0 s, differenced per asset (fleet total last reported: 7581, at the first tick: 0); this ranks when the energy went, not how much of it there was
accuracy: not measured (no truth in log)
    no episode log carries truth targets yet, so there is nothing to measure against
total: 0.9916 (weighted over the 2 of 5 axes this log answers for, 400 records)
ARENA.md section 5 names seven criteria; the five above are the ones a log can answer for. Autonomy and collaboration are read off the fleet's behaviour and the explanation, not off a log, so they are not axes here.
```

The gate's smoke log (`run --seed 7 --ticks 400`) prints the same numbers, which is expected: `fixtures/episodes/demo.jsonl` is that run.

Acceptance criteria, and the test that covers each:

| criterion | test |
|---|---|
| a real value for every axis it can compute | `tests/test_score.py::test_demo_episode_scores_non_zero_bounded_numbers` — coverage and search efficiency over `demo.jsonl`, asserted inside `(0, 1]` |
| an axis it cannot compute says so in words | `::test_demo_episode_reports_the_axes_it_cannot_answer_for_in_words`, `::test_search_efficiency_says_so_when_no_energy_was_spent`, `::test_accuracy_is_never_a_number`, `tests/test_cli.py::test_score_never_prints_a_zero_for_an_axis_it_did_not_measure` |
| each axis names what it was derived from | `::test_demo_episode_names_a_source_for_every_axis`, `tests/test_cli.py::test_score_shows_where_each_axis_came_from` |
| `--weights` still works, total uses only computed axes | `::test_the_total_weighs_only_the_axes_the_log_answered_for`, `::test_a_weight_can_silence_an_axis_without_dragging_the_total_down`, `::test_the_committed_weights_file_carries_exactly_the_axes` |
| a test over `fixtures/episodes/demo.jsonl` | the four `test_demo_episode_*` tests above |
| `AXES` reconciled with `ARENA.md` §5 | `::test_the_two_criteria_no_log_can_answer_for_are_named` |

Per-axis behaviour is pinned by value, not only by ordering, so a metric that silently changes shape fails: `::test_coverage_is_the_last_tick_not_the_peak`, `::test_detecting_sooner_scores_higher_than_detecting_later`, `::test_tracking_duration_is_the_share_of_ticks_holding_a_contact`, `::test_search_efficiency_rewards_covering_before_spending`. The two ordering tests previously asserted only an ordering plus a substring, which two shape-changing mutations survived (a squared detection-speed curve, and crediting each energy delta to the previous tick's coverage); both now carry a `pytest.approx` on the expected number, and both mutations fail them.

The round-1 review's other findings are pinned too: `::test_a_lost_contact_is_not_a_tracked_tick` and `::test_a_target_detected_and_never_held_scores_a_measured_zero` (the lost-contact axis), `::test_demo_episode_does_not_report_a_tracking_duration_it_never_measured` (no fake zero for an absence), `::test_search_efficiency_differences_energy_per_asset` (a silent asset), `tests/test_cli.py::test_score_refuses_a_negative_or_non_finite_weight` (a bad weights file printed `total: 0.0000` or `total: nan`), and `tests/test_gate_config.py::test_the_smoke_score_check_refuses_words_where_a_number_is_owed` (the gate accepted any non-numeric string for four of the five axes).

`demo.jsonl` carries no contacts, so detection speed and tracking duration are exercised against hand-built episodes rather than against it.

## Gate

`python scripts/gate.py` fails fast at step 4 on a pre-existing error that is not from this branch, so the steps after it were run by hand with the gate's own commands and the gate's own interpreter (`.venv/Scripts/python.exe`). Verbatim:

```
GATE PASS: install (11.6s)
GATE PASS: lint (0.6s)
GATE PASS: format (0.6s)
whiteout\vision\imagery.py:267: error: Unused "type: ignore" comment  [unused-ignore]
Found 1 error in 1 file (checked 38 source files)
GATE FAIL: typecheck (15.2s) -- exit code 1
```

That typecheck error reproduces on unmodified `main` — it is a local venv exposing a system `cv2`, nothing in this diff touches `whiteout/vision/imagery.py`, and CI does not hit it. The remaining four steps, run separately:

```
956 passed, 12 skipped, 1 deselected in 57.75s          # test: pytest -q -m "not slow"
Successfully built whiteout-0.1.0-py3-none-any.whl      # build: python -m build --wheel --no-isolation
smoke: None                                             # gate.step_smoke(), no failure reason
  logs are byte-identical (1242066 bytes)
determinism: None                                       # gate.step_determinism(), no failure reason
```

`smoke` and `determinism` were run through `scripts/gate.py`'s own `step_smoke` and `step_determinism` functions, so the changed smoke check is what passed, not a hand-written substitute.

Re-run after the round-1 fixes and after merging `main` (`970 passed, 12 skipped, 1 deselected`), with the same one pre-existing typecheck failure and every other step through the gate's own step functions.

## Follow-ups

- `hackathon/SPEC.md` §2's demo script still says "four score dials" and names the old four axes in beats 1, 2 and 7, and `viz/viewer.js:41-46` still declares the old four (`coverage`, `collaboration`, `efficiency`, `tracking_accuracy`), with no dial for `detection_speed` or `tracking_duration`. Nothing breaks — `renderDials` only ever sets `is-unknown` — but the viewer and the demo script need the same reconciliation this PR made in §3, §5 and §6.
- `search_efficiency` is scale-invariant in energy by construction, so it does not penalise a fleet that burns ten times the energy on the identical coverage curve. The docstring, the derivation line and this body now say that plainly rather than implying thrift, but whether the axis should be renamed (it is a promptness-of-spend measure) or divided by something that grows with total energy is a product decision, not a review fix. Left open.
- `TrackHold._last` is never cleared, so `Coordinator` emits a `state="lost"` contact on every tick after the first sighting for the rest of the run. The scorer no longer counts those as tracked ticks, which is the in-scope fix; whether the track hold should stop emitting a contact once it is lost is a change to its lifecycle and belongs to its own ticket.
- `SPEC.md` §6's smoke row and `scripts/gate.py` are the only harness files this touches, and only because the axis list moved. Flagging it because `CLAUDE.md` holds harness changes for a human merge.
- Accuracy stays unmeasurable until something populates `Truth.targets` during an arena run. That is not #26 (the scoring model) and has no ticket that I could find.
- `fixtures/episodes/demo.jsonl` carries no contacts and no truth, so two of the five axes cannot be demonstrated against the committed episode. An episode fixture that holds a contact would make the demo output show four numbers rather than three.


Ollienel777:
cold: findings @ 6302b36 — 1 Critical, 7 Medium, 6 Nit. `tracking_duration` counts a `lost` contact as a tracked tick, so it will read ~0.97 for a target lost on sight; it also prints the `0.0000` #130 exists to remove, and the relaxed gate smoke check no longer proves any axis but `coverage` is a number. Detail in the review.

Ollienel777 (commented):
cold: findings @ 6302b36 — 1 Critical, 7 Medium, 6 Nit. Round k=1.

Reviewed at detached `6302b36`. `/code-review high 133` ran and named files that are in this PR's diff (`whiteout/score/episode.py`, `whiteout/cli.py`, `scripts/gate.py`, `fixtures/weights/equal.json`, `hackathon/SPEC.md`, `tests/test_score.py`, `tests/test_cli.py`), so its scope was correct; every finding below I re-derived and anchored myself.

What I ran: `whiteout score fixtures/episodes/demo.jsonl --weights fixtures/weights/equal.json` (output matches the PR body verbatim); `pytest -q tests/test_score.py tests/test_cli.py tests/test_gate_config.py` (64 passed); `ruff check .` (clean); two mutations of `whiteout/score/episode.py`; direct calls into `scripts/gate.py:_score_failure`; weights files with a negative and a `NaN` weight. The known `imagery.py:267` typecheck failure is excluded as pre-existing.

---

## Critical

### R1-C1 — `tracking_duration` counts a **lost** contact as a tracked tick

`whiteout/score/episode.py:161-162`

```python
def _contact_ticks(records: Sequence[EpisodeRecord]) -> list[int]:
    return [index for index, record in enumerate(records) if record.contacts]
```

The tuple being non-empty is not the same statement as "we are holding it".

- `whiteout/tracks/maintain.py:174-184`: `TrackHold.state` returns `"lost"` once `self._now - self._last.t > coast_s`. `self._last` is only ever *assigned* a new sighting — it is never cleared, so once a sighting has landed it stays set for the rest of the run.
- `whiteout/coordinate.py:428-436`: `_contacts` returns `()` only when `self._hold is None or self._hold.last_sighting is None`. Otherwise it emits a `Contact`, mapping `"lost" -> "lost"`.

So after the first sighting, **every** subsequent record carries a non-empty `contacts`, whatever the track's state. `tracking_duration` therefore measures *time since first detection*, not time tracking.

Concrete: the fleet glimpses the vessel once at tick 10 of the 400-tick arena run and never sees it again. 390 records carry a `state="lost"` contact, and the demo screen prints `tracking_duration: 0.9750` — a near-perfect tracking score for a target lost on sight — and that number lifts the weighted total a judge reads. The derivation line (`390 of 400 ticks carry a contact`) is literally true and gives the reader no way to notice.

Issue #130's acceptance text does say "ticks with a non-empty `contacts`, over total ticks", so this follows the ticket to the letter. The ticket is wrong on this point and the data to do better is right there: filter on the states that mean we are holding it (`tracked`, and `handed_off` if the lifecycle emits it) rather than on the tuple being non-empty. `demo.jsonl` carries no contacts, so this is latent today and fires the moment the vision path that landed in #110/#111/#114 produces one.

---

## Medium

### R1-M1 — `tracking_duration` prints the exact fake zero #130 was filed to remove

`whiteout/score/episode.py:196-198`, output on `fixtures/episodes/demo.jsonl`

`_detection_speed` and `_tracking_duration` read the same `_contact_ticks` result. When it is empty, one reports `not detected` in words and is dropped from the total; the other reports a hard number:

```
detection_speed: not detected
    no record of 400 carries a contact, so there is no first detection to time
tracking_duration: 0.0000
    0 of 400 ticks carry a contact (0.0 s of 199.5 s)
```

That inconsistency is not free. `coverage: 1.0000` and `search_efficiency: 0.9833` would give a total of `0.9917`; the `0.0000` drags it to the `0.6611` the command prints. So the single most prominent number on the demo screen is a third lower because of an axis reporting the same absence its sibling reports in words.

This is also the committed smoke path: the gate's `run --seed 7 --ticks 400` **is** `demo.jsonl`, so the gate's own smoke output carries the zero too. Acceptance criterion 2 of #130 — "an axis that cannot be computed says so in words, never `0.0000`" — is met for three axes and not for this one.

### R1-M2 — the smoke check no longer proves the scorer computes anything but coverage

`scripts/gate.py:502-521`

`_score_failure` requires only that a line exists for every axis, that `coverage` parses as a finite float, and that any *other* axis which happens to parse is finite. Line 518 accepts any non-numeric string unconditionally. I called it directly:

```
real                                      -> None
search_eff -> words, coverage -> 0.0000   -> None
every axis words, coverage: 0.0000        -> None
coverage: nan                             -> 'score printed a non-finite coverage: ...'
```

So a regression that makes four of the five axes print words — a stub, an energy-plumbing regression putting `search_efficiency` back to `no energy recorded`, a lifecycle regression putting `tracking_duration` and `detection_speed` back to words — passes the gate green while `coverage` is a number. That is precisely the class of failure #130 exists to catch.

The one remaining numeric arm is also unreachable: `_coverage` (`episode.py:152`) routes through `_clamp` (`episode.py:142-145`), which maps every non-finite value to `0.0` and clamps into `[0, 1]`, and `f"{value:.4f}"` of a finite float always parses. So `ALWAYS_NUMERIC`'s finiteness check cannot fail, and its comment at `gate.py:71-73` ("a run that cannot print this one as a finite number is broken") describes a branch no run can reach. The check that *can* fire is "an axis line is missing", nothing more.

Pin the accepted words to the set the scorer can actually emit (`not detected`, `no energy recorded`, `not measured (no truth in log)`) so an unexpected phrase fails, and consider requiring `search_efficiency` to be a number on the smoke run, which by construction it always is.

### R1-M3 — `_search_efficiency` differences the fleet-wide sum, so a silent asset mis-attributes its whole energy

`whiteout/score/episode.py:215-222`

```python
previous = sum(pose.energy_used for pose in records[0].observation.poses)
...
    total = sum(pose.energy_used for pose in record.observation.poses)
    delta = max(0.0, total - previous)
```

`ArenaTransport.observe` (`whiteout/transport/arena.py:345-350`) deliberately omits a pose for an asset that has not reported this tick — "Silence is not a fault: an asset that has not reported yet simply contributes no pose this tick." When an asset drops out, `total` falls, `max(0.0, ...)` clamps those ticks to zero, and when it reports again its *entire* cumulative `energy_used` arrives as one delta weighted by that one tick's `covered_fraction`. On a live arena run with a flaky link the axis can swing either way by an arbitrary amount.

The kinematic transport always emits every pose, so the smoke run and `demo.jsonl` never show this — it is an arena-day defect only. Difference `energy_used` per `asset_id` and sum the per-asset increments.

### R1-M4 — weights are not validated, and a bad weights file prints a fake zero or `nan`

`whiteout/cli.py:224-229`, `whiteout/score/episode.py:117-129`

`cmd_score` accepts anything `float()` takes; `weighted_total` guards only `divisor <= 0.0`, returning `0.0`. Both run against the committed demo log:

```
--weights {"coverage": -4.0, others 1.0}  ->  total: 0.0000 (weighted over the 3 of 5 axes ...)
--weights {"coverage": NaN,  others 1.0}  ->  total: nan    (weighted over the 3 of 5 axes ...)
```

Python's `json` accepts the bare token `NaN`, so the second is reachable from a plain file. The first is the worst case on the demo path: a `0.0000` total that is not a measurement at all, printed with full confidence beside five derivation lines — the exact output shape this PR exists to abolish. `--weights` is the tuner's input, so a sweep that emits a negative weight silently scores every episode zero. The gate does not look at the `total:` line, so neither is caught. Reject non-finite and negative weights at load, and say which axis.

### R1-M5 — the PR body's claim that the per-axis tests pin each metric's shape is false

PR body: *"Per-axis behaviour is pinned separately, so a metric that silently changes shape fails."*

Two mutations, each made to change the metric's meaning and nothing else, both leave `tests/test_score.py` and `tests/test_cli.py` fully green (32 passed):

1. `whiteout/score/episode.py:222` — credit each energy delta to the *previous* tick's coverage instead of the current one (an off-by-one that changes the metric from "coverage bought by this spend" to "coverage before this spend"). All 32 pass; the demo output moves from `search_efficiency: 0.9833` to `0.9811`.
2. `whiteout/score/episode.py:184` — `1.0 - elapsed / span` becomes `1.0 - (elapsed / span) ** 2`, a different detection-speed curve. All 32 pass.

`test_search_efficiency_rewards_covering_before_spending` (`tests/test_score.py:197-212`) and `test_detecting_sooner_scores_higher_than_detecting_later` (`:183-188`) assert only an ordering plus a substring; neither pins a value. `test_coverage_is_the_last_tick_not_the_peak` and `test_tracking_duration_is_the_share_of_ticks_holding_a_contact` *do* pin exact values — so the fix is small: one `pytest.approx` on the expected number in each of the two ordering tests.

### R1-M6 — `hackathon/SPEC.md` now contradicts itself, and `docs/arena.md` is made false by this diff

`hackathon/SPEC.md:187` (this diff) vs `hackathon/SPEC.md:131`; `docs/arena.md:178`

- `SPEC.md:187` now says the scorer's model is "coverage, detection speed, tracking duration, search efficiency and accuracy — five of `ARENA.md` §5's seven criteria". `SPEC.md:131`, in §3's criteria table, still says "coverage, collaboration, efficiency, tracking accuracy | Our reimplementation of **all four** with weights as parameters; the coordinator optimises them directly". As written, the spec now asserts two different scorer models. CLAUDE.md makes the spec the authority for the gate, so this matters beyond tidiness. The PR body's follow-up names §2's demo script (beats 1, 2, 7) but not §3's table row.
- `docs/arena.md:178` says "`score` prints the **four** axes this repository computes. They are **not** the seven the sponsor judges on". This diff makes the first half false (it prints five) and the second half half-false (five of them now *are* five of the seven, which is the reconciliation this PR added). `docs/arena.md:175` is the arena-day runbook line that invokes the very command this PR changed, so it is read on the day.

Neither file is in the diff; both are made wrong by it. `docs/` and `SPEC.md` are harness files a human merges, so flagging rather than fixing is right — but the follow-up list should name them.

### R1-M7 — `search_efficiency` does not penalise energy at all, which its one-line gloss and #130 both imply it does

`whiteout/score/episode.py:211-242`; module docstring lines 34-42

The docstring's headline is "Coverage measured against the energy that bought it", and #130 asked for "total `observation.poses[].energy_used` at the last tick, against coverage reached". What landed is a *convex combination*: `sum(cov_i * delta_i) / sum(delta_i)`. That is scale-invariant in energy — the magnitude cancels. Four episodes, same coverage trajectory:

```
A  10 units/tick                 : 1.0
B 100 units/tick, same coverage  : 1.0
C idles burning 8010 extra units : 1.0
D shuts down after covering      : 1.0
```

A fleet burning ten times the energy for the identical coverage curve scores identically, and a fleet idling at full coverage burning energy forever scores 1.0 — it can only *raise* the score, never lower it. The full paragraph at docstring lines 36-42 does spell the formula out honestly, and the "Bounded in `[0, 1]`, unlike a bare coverage-per-joule ratio" argument is sound. But the axis is named `search_efficiency`, the gloss says "against the energy that bought it", and the number on the demo screen is `0.9833` next to 7581 units spent. A judge reads that as "we were efficient with our energy", which this metric never measured. Either rename the axis to what it measures (coverage-weighted promptness of spend) or divide by something that grows with total energy.

---

## Nit

- **R1-N1** — `whiteout/score/episode.py:184`: `promptness = 1.0 if span <= 0.0 else ...` gives a perfect `1.0000` to a single-record log, or to one whose records share a `t`. That is a measured-looking number with no measurement behind it; the sibling axes say so in words in the equivalent case.
- **R1-N2** — the demo output has a 150-character line (`search_efficiency`'s derivation) and a 215-character closing line. On a demo screen at a legible font both wrap to two or three lines, turning a 12-line block into ~17 and breaking the axis / indented-derivation rhythm that makes the output scannable. Wrapping to ~78 columns would keep it.
- **R1-N3** — `viz/viewer.js:41-46` still declares the old four (`coverage`, `collaboration`, `efficiency`, `tracking_accuracy`), so the dial row in beat 1 is labelled with axes the scorer no longer produces and has no dial for `detection_speed` or `tracking_duration`. Nothing breaks — `renderDials` (`viewer.js:331-343`) only ever sets `is-unknown` — and the PR body does say "the viewer and the demo script need the same reconciliation", so this is the acknowledged follow-up, recorded here with the line number.
- **R1-N4** — `scripts/gate.py:57-58` says its `AXES` is "kept in step with `whiteout.score.AXES`", and nothing enforces it. Divergence by rename fails loudly (a missing axis line), but an axis *added* to `whiteout.score.AXES` is silently unchecked by the gate. `tests/test_gate_config.py` already imports the gate module (`_load_gate`, line 46) — one assert would pin it.
- **R1-N5** — `whiteout/cli.py:13` is 92 characters in a docstring the rest of the file wraps at ~78; `ruff format` does not reflow docstrings, so it stays ragged.
- **R1-N6** — `whiteout/score/episode.py:206`: the seconds in `tracking_duration`'s derivation are `share * span`, which assumes evenly spaced ticks. True for the current fixed-tick clock; it will quietly stop being the held time if ticks ever vary. Summing the intervals of the held ticks costs the same line.

---

## Design (not a finding)

**Does it solve #130's problem?** Yes, and the shape is right. The problem was four constants on a judge's screen; the answer is a pure function of the log, a named derivation under every axis, and words instead of a zero where the log cannot answer. Putting it in `whiteout/score/` (where `SPEC.md` §5 already assigned it) rather than inline in `cli.py` is correct, and `cmd_score` reduced to read-derive-print is the right division. The `AxisScore(value | words | derivation)` triple is a good primitive: it makes "unmeasured" a first-class state rather than a sentinel number, and `weighted_total` dropping unmeasured axes from numerator *and* denominator is the right arithmetic. The `unscorable_criteria()` line answers #130's last criterion honestly instead of quietly shipping four axes.

**Is there a simpler shape?** Not materially — the module is about as small as five axes and their prose can be. One simplification worth taking: `_detection_speed` and `_tracking_duration` both branch on `_contact_ticks` being empty and disagree about what that means (R1-M1). Deciding once — "no contact in the log" is a single unanswerable condition for both axes — would remove the inconsistency, remove the `0.0000`, and make R1-C1's fix a change in one predicate rather than two.

**The honesty question on the committed fixture.** The output is honest as far as it goes: nothing is invented, every line names its source, and `not detected` / `not measured (no truth in log)` are exactly right. But taken as a whole, `demo.jsonl` produces `coverage: 1.0000`, `search_efficiency: 0.9833`, `detection_speed: not detected` — a fleet that swept 100% of the strait and found nothing, scoring 0.98 on efficiency. A judge who reads the axes in order gets "near-perfect" from the top two lines and "we never found it" from the third, and the two do not reconcile. The PR's own follow-up already names the cure: an episode fixture that holds a contact. Until then, the demo is better served by a fixture where the numbers tell one story than by four honest numbers telling two.

Ollienel777:
loop: fixer dispatched @ 6302b36 — findings

R1-C1 plus the Mediums raised in round 1.

Ollienel777:
fix: pushed @ 6086cbb

`main` moved under this branch while the fixes were being made (`85154d6`), so the push carries a merge of it as well. `hackathon/SPEC.md` §5 conflicted; see R1-M6.

R1-C1: fixed — in the scorer, not in the track hold. `tracking_duration` now counts only ticks carrying a contact in `HELD_STATES` (`tracked`, `handed_off`); `detection_speed` still reads the first record carrying any contact, because that is a detection whatever its state. `test_a_lost_contact_is_not_a_tracked_tick` builds the episode the review describes — two held ticks, then sixteen `state="lost"` ones — and pins the axis at 2/20 rather than 18/20. The observation about the source stands and is not fixed here: `TrackHold._last` is never cleared, so `Coordinator._contacts` emits a `lost` contact on every tick after the first sighting for the rest of the run. That is the track hold's lifecycle, outside #130, and it is now in the PR body's follow-ups.

R1-M1: fixed. A log carrying no contact at all is one unanswerable condition, decided once: `tracking_duration` reports `not detected` in words, as `detection_speed` does, and is dropped from the total. A zero there is now a measurement — contacts in the log, none ever held — and `test_a_target_detected_and_never_held_scores_a_measured_zero` pins that case separately. On `demo.jsonl` the total moves from `0.6611` to `0.9916`.

R1-M2: fixed. `ALWAYS_NUMERIC` is now `("coverage", "search_efficiency")`, and every other axis must be either a finite number or one of `ACCEPTED_WORDS` — the exact three phrases the scorer can emit. An axis printing anything else fails with the string it printed. `test_the_smoke_score_check_refuses_words_where_a_number_is_owed` calls `_score_failure` directly with the real output, a four-axes-in-words stub, an invented phrase and a `nan`.

R1-M3: fixed. `_search_efficiency` differences `energy_used` per `asset_id` and sums the per-asset increments. `test_search_efficiency_differences_energy_per_asset` runs the same episode twice, once with an asset silent for two ticks: the fleet-wide difference scored it 0.9167, per-asset scores it 1/3, the same as the episode where nothing drops out.

R1-M4: fixed. `cmd_score` rejects a non-finite or negative weight at load and names the axis, so `total: 0.0000` and `total: nan` are no longer reachable from a weights file. `test_score_refuses_a_negative_or_non_finite_weight` covers both, including the bare `NaN` token plain JSON accepts.

R1-M5: fixed, tests first. `test_detecting_sooner_scores_higher_than_detecting_later` and `test_search_efficiency_rewards_covering_before_spending` now pin values with `pytest.approx` (`1 - 1/9` and `1 - 8/9`; `1.0` and `10/90`) alongside the ordering. Both named mutations — the squared detection-speed curve, and crediting each energy delta to the previous tick's coverage — fail them. The body's sentence is amended to say "pinned by value, not only by ordering" and to name what the two tests previously missed.

R1-M6: fixed, and smaller in one place than the review implies. `ee35f1c` on `main` had already rewritten §5's component table, deleting the `Scorer` row and adding a paragraph stating that `whiteout/score/` **does not exist**, which this branch falsifies. The merge takes `main`'s table and puts back a `Scorer` row and a corrected `CLI` row, and rewrites that paragraph: the package exists and reads the log, and what has not changed is that nothing optimises against it. §3's criteria-table row (`all four`, `the coordinator optimises them directly`) now points at §5's axis list and says the scorer is read as an instrument, not optimised against. `docs/arena.md:178` says five axes and which five. `README.md`'s honesty section, new on `main`, said the scorer prints four zeroes without reading the log; corrected. §2's demo script and `viz/viewer.js` still name the old four and remain the follow-up, now spelled out in the body.

R1-M7: declined, needs a decision. Renaming the axis changes a key in `fixtures/weights/*.json`, `scripts/gate.py`, `SPEC.md` and the viewer; dividing by total energy changes what the number means and gives up the `[0, 1]` bound the PR argued for. Both are product choices about what the judge's screen should measure, not review fixes, and it is four hours to submit. What is fixed is the false claim around it: the module docstring, `_search_efficiency`'s docstring, the printed derivation line and the PR body now all say the axis is scale-invariant in energy and ranks when the spend went rather than how large it was. The choice is recorded in the body's follow-ups.

Nits R1-N1 through R1-N6 not attempted — outside the set this pass was given.

Gate, verbatim:

```
GATE PASS: install (0.7s)
GATE PASS: lint (0.4s)
GATE PASS: format (0.3s)
whiteout\vision\imagery.py:267: error: Unused "type: ignore" comment  [unused-ignore]
Found 1 error in 1 file (checked 38 source files)
GATE FAIL: typecheck (2.5s) -- exit code 1
```

That typecheck failure is the known pre-existing one on `whiteout/vision/imagery.py:267`, which reproduces on unmodified `main` and which nothing in this branch touches. The four steps after it were run through `scripts/gate.py`'s own `step_test`, `step_build`, `step_smoke` and `step_determinism` functions on the merged tree, so the changed smoke check is what ran:

```
step_test: None          # 970 passed, 12 skipped, 1 deselected in 54.91s
step_build: None         # Successfully built whiteout-0.1.0-py3-none-any.whl
step_smoke: None
step_determinism: None   # logs are byte-identical (1242066 bytes)
```

Ollienel777 (commented):
semi-cold: does not close @ 6086cbb — 6 closed, 1 withdrawn, 1 open, 1 new Medium. Round k=2.

Reviewed at detached `6086cbb`. Every verdict below was taken against the code at that SHA, not against the fixer's reply: I re-ran the two mutations R1-M5 names, added a third of my own for R1-C1, called `scripts/gate.py:_score_failure` directly with five inputs the fixer does not name, ran `cmd_score` against three weights files, and grepped the tree for every surviving copy of the old four-axis claim. Full suite on the merged tree: `970 passed, 12 skipped, 1 deselected`; `ruff check .` clean; `ruff format --check whiteout tests scripts` clean. The known `imagery.py:267` typecheck failure is excluded as pre-existing.

I separated the fixer's commit (`3d24b7b`) from the merge of `main` (`85154d6`); against `main` this PR's whole surface is `README.md`, `docs/arena.md`, `fixtures/weights/equal.json`, `hackathon/SPEC.md`, `scripts/gate.py`, `whiteout/cli.py`, `whiteout/score/` and three test files.

---

## Verdicts

**R1-C1 — `tracking_duration` counts a lost contact — closed.**
`whiteout/score/episode.py:188-194` filters on `HELD_STATES` (`:91`), and `whiteout/coordinate.py:430-436` is the mapping that makes that the right predicate: `held`/`coasting` to `"tracked"`, `lost` to `"lost"`, `unseen` to `"unconfirmed"`. So the states the scorer counts are exactly the ones the source emits while holding. I did not take the named test on trust: I reverted `episode.py:193` to `if record.contacts` and ran `tests/test_score.py` — `test_a_lost_contact_is_not_a_tracked_tick` **and** `test_a_target_detected_and_never_held_scores_a_measured_zero` both fail, then restored the file. The scorer-side fix genuinely closes it rather than hiding it: `_contact_ticks` (`:183-185`) is now used only by `detection_speed`, where "any contact, whatever its state" is the correct predicate for timing a first detection. The source problem (`TrackHold._last` never cleared) is a lifecycle change with other consumers and is correctly an out-of-scope follow-up.

**R1-M1 — the `0.0000` for an absence — closed.**
`whiteout/score/episode.py:228-238` returns words for a log with no contact anywhere, and the run confirms it:

```
tracking_duration: not detected
    no record of 400 carries a contact, so there was never anything to hold
total: 0.9916 (weighted over the 2 of 5 axes this log answers for, 400 records)
```

Deciding the absence once rather than twice is the simplification round 1's design note asked for.

**R1-M2 — the smoke check accepted any string — closed.**
I loaded `scripts/gate.py` and called `_score_failure` directly. The four the fixer names behave as claimed (real to `None`; four-axes-in-words stub refused on `search_efficiency`; `'pending'` refused, naming the string; `nan` refused). Five the fixer does **not** name:

| input | `_score_failure` |
|---|---|
| `search_efficiency: no energy recorded`, everything else honest | refused — `"non-numeric search_efficiency"` |
| `coverage: inf` | refused |
| a missing axis line | refused |
| `coverage: 2.5` (outside `[0, 1]`) | `None` — see R2-N1 |
| five `0.0000` lines (the stub #130 removed) | `None` — see R2-N1 |

`ALWAYS_NUMERIC` at `gate.py:77` and `ACCEPTED_WORDS` at `:84` are what round 1 asked for, and both requested changes landed.

**R1-M3 — `_search_efficiency` differences the fleet-wide sum — open.**
The scenario the fixer's test builds is fixed, and the per-asset dict at `episode.py:268-270` is the right shape. But the seeding is only from `records[0]`, so `episode.py:276` — `increment = pose.energy_used if before is None else pose.energy_used - before` — still credits an asset's **whole cumulative `energy_used`** to the first tick it appears on, if it did not appear at tick 0. That is the same mis-attribution, on the same `ArenaTransport.observe` behaviour: an asset that has not reported contributes no pose, and tick 0 is precisely when the fewest assets have reported. Three runs of the same episode, coverage 0 until the last tick, quad spending 10 units a tick, rover holding a pre-episode 100 units and spending nothing:

```
rover present throughout   -> 0.3333   (30 units differenced)
rover silent on ticks 1, 2 -> 0.3333   (30 units — the fix works)
rover absent at tick 0     -> 0.0769   (130 units: its pre-episode burn lands on tick 1)
```

The third is not a hypothetical variant of the finding, it is the finding: energy the fleet did not spend during the episode is attributed to one tick's coverage and swings the axis by a factor of four. Treat the first sighting of an `asset_id` as a baseline, exactly as tick 0 is treated, and add the case to `test_search_efficiency_differences_energy_per_asset` (`tests/test_score.py:246-282`), which today covers only the drop-out-and-return shape.

**R1-M4 — weights not validated — closed.**
`whiteout/cli.py:234-247`. Both cases run against the committed demo log exit 1 and name the axis: `score: weights file has a negative weight for coverage: -4.0`, `score: weights file has a non-finite weight for coverage`. One residual I am not reopening on: an all-zero weights file is non-negative and finite, so it still reaches `weighted_total`'s `divisor <= 0.0` arm (`episode.py:149-150`) and prints `total: 0.0000` at exit 0. Degenerate input, but if a sweep can emit a negative weight it can emit zeros; worth a line in a later pass.

**R1-M5 — the PR body's "pinned separately" claim — closed.**
I re-ran both named mutations rather than taking the claim. Squaring the detection curve at `episode.py:216` gives `FAILED tests/test_score.py::test_detecting_sooner_scores_higher_than_detecting_later`. Crediting each energy delta to the previous tick's coverage at `episode.py:280` gives `FAILED ::test_search_efficiency_rewards_covering_before_spending` and `FAILED ::test_search_efficiency_differences_energy_per_asset`. Both restored afterwards; the tree is clean. The `pytest.approx` values at `tests/test_score.py:222-226` and `:285-291` are what does it, and the body's sentence is amended to "pinned by value, not only by ordering", which is now true.

**R1-M6 — the false four-axis claims — closed.**
Grepped the tree at `6086cbb` for every surviving copy. `hackathon/SPEC.md:131` now defers to section 5's axis list and — after the merge, not in `3d24b7b` — says "read as an instrument rather than optimised against", which the fixer's own commit had still contradicted; the merge resolution is the better text. `SPEC.md:209` puts back a `Scorer` row and corrects `main`'s paragraph claiming `whiteout/score/` does not exist. `docs/arena.md:180-186` says five and names them. `README.md:174-181` replaces "the scorer is a stub ... prints four zeroes". `fixtures/weights/equal.json` carries the five keys. The two left as follow-ups are acceptable: `SPEC.md:88` (beat 1's narration) was already contradicted by `hackathon/ARENA.md:132` before this diff touched anything, so it is not made wrong here; and `viz/viewer.js:42-46` is inert, since `renderDials` only ever sets `is-unknown`. Both are named in the body. The other hits are quotations of the sponsor's prize text (`SPEC.md:120`, `hackathon/tracks/prizes.md:158`), a historical narration (`whiteout/transport/kinematic.py:15`) and a leak-guard test's vocabulary (`tests/test_transport.py:339-345`) — none is a claim about what `score` prints.

**R1-M7 — `search_efficiency` does not penalise energy — withdrawn.**
I judged this on its merits rather than on the decline. The harm round 1 identified was a false claim — the gloss saying "against the energy that bought it" — and that is fixed in all three places a reader meets it: the module docstring (`episode.py:42-55`), the function docstring (`:255-259`) and, decisively, the line printed on the judge's screen, which now ends `this ranks when the energy went, not how much of it there was`. Nothing on the demo screen now reads as a thrift claim. The residue is the two remedies round 1 offered, and both are genuinely product choices: `search_efficiency` is the sponsor's own section 5 criterion name, so renaming it breaks the mapping this PR just built, and dividing by total energy gives up the `[0, 1]` bound. Recorded in the follow-ups, not blocking. This is a decline I agree with, not a review fix dressed as one — but note that it rests entirely on the printed caveat, so deleting that derivation clause reopens the finding.

---

## New

### R2-M1 (Medium) — `SPEC.md` section 6 and the PR body both understate the gate check the fixer just tightened

`hackathon/SPEC.md:398` vs `scripts/gate.py:77`

`3d24b7b` widened `ALWAYS_NUMERIC` to `("coverage", "search_efficiency")` — correctly, that is R1-M2's fix — but the spec row that defines the step still reads:

> ... each one either a finite number or the words saying why this log cannot answer for it (#130); `coverage` is always a number

and the PR body's "What" section still says the check "requires `coverage` to be a finite number ... and accepts words for the rest", which is the pre-fix behaviour and is the exact weakness R1-M2 was raised about. Two documents now describe a gate weaker than the one in the tree.

This matters twice over. `CLAUDE.md` makes `hackathon/SPEC.md` the authority for the gate, so a later hand reconciling them has a spec telling them to loosen `gate.py`. And the body is squash-merged onto `main`, so the wrong description becomes the permanent record of this change. Both are one-line edits: name both axes in `SPEC.md:398`, and rewrite the body sentence to say words are pinned to the three phrases the scorer can emit. Raised at Medium for consistency with round 1, which graded a false PR-body claim (R1-M5) and a spec contradiction (R1-M6) at the same level.

### R2-N1 (Nit) — the gate's smoke check still passes the stub this PR removed

`scripts/gate.py:502-521`. Verified directly: `_score_failure` returns `None` for output printing `0.0000` on all five axes — the shape `scores = {axis: 0.0 for axis in AXES}` produced, which is what #130 exists to abolish. It also passes `coverage: 2.5`, though `_clamp` (`episode.py:164-167`) makes that unreachable today. Not raised higher because `tests/test_score.py`'s pinned demo values fail first, so `step_test` catches the regression before `step_smoke` sees it — the smoke step is the second line, not the first. A `0.0 < coverage <= 1.0` bound on `ALWAYS_NUMERIC` would close both.

### R2-N2 (Nit, on the demo screen) — the fix moved the committed episode's headline number from 0.66 to 0.99

`fixtures/episodes/demo.jsonl` now scores `total: 0.9916`, up from `0.6611`, because `tracking_duration` correctly left the denominator. The arithmetic is right and the change is the point of R1-M1. But the screen a judge reads in beat 2 now says `0.9916` above `detection_speed: not detected` — a near-perfect total for a run that never found the ship. Round 1 raised this as a design note when the number was 0.66; at 0.99 it is sharper, and the cure is already in the body's last follow-up (an episode fixture that holds a contact). Worth pulling forward if anything is left before submit, because it is the one number on that screen a reader takes without its derivation line.

---

## What I did not re-derive

The six round-1 nits (R1-N1 through R1-N6) were outside the set this round was given and I did not re-check them; R1-N3 (`viz/viewer.js`) I confirmed only as far as R1-M6 needed. I ran the gate's `lint`, `format` and `test` steps and read `step_smoke`'s check by calling it directly, but did not run `step_smoke` or `step_determinism` end to end — the fixer's report of those two is unverified by me.

Ollienel777:
semi-cold: does not close @ 6086cbb — 6 closed, 1 withdrawn, 1 open, 1 new Medium. Round k=2.

R1-C1 closed, R1-M1 closed, R1-M2 closed, R1-M3 **open**, R1-M4 closed, R1-M5 closed, R1-M6 closed, R1-M7 **withdrawn**. New: R2-M1 (Medium), R2-N1, R2-N2.

Two things hold it. **R1-M3 is only half fixed**: `_search_efficiency` now differences per `asset_id`, but `whiteout/score/episode.py:276` seeds the baseline from `records[0]` alone, so an asset that did not report on tick 0 still dumps its whole cumulative `energy_used` onto the first tick it appears on. The same episode scores 0.3333 with the rover present throughout, 0.3333 with it silent mid-run (the fix works), and 0.0769 with it merely absent at tick 0. Tick 0 is exactly when the fewest assets have reported on an arena run. **R2-M1**: `3d24b7b` widened `ALWAYS_NUMERIC` to `("coverage", "search_efficiency")`, but `hackathon/SPEC.md:398` and the PR body still describe the pre-fix check — and `CLAUDE.md` makes the spec the gate's authority, so the divergence points the wrong way.

What I checked rather than took on trust: I re-ran both mutations R1-M5 names (squared detection curve, previous-tick coverage) and both fail; I added a third for R1-C1 (revert `episode.py:193` to `if record.contacts`) and two tests fail; I called `scripts/gate.py:_score_failure` directly on the fixer's four cases and five more it does not name; I ran `cmd_score` against negative, `NaN` and all-zero weights files; and I grepped the whole tree for surviving copies of the four-axis claim. `970 passed, 12 skipped, 1 deselected`, `ruff check .` and `ruff format --check whiteout tests scripts` clean, the `imagery.py:267` typecheck failure excluded as pre-existing. The working tree was restored after every mutation.

A semi-cold round does not settle. Verdict lines and anchors are in the review.

Ollienel777:
loop: the operator has asked for R1-M3 to be fixed rather than recorded.

`FIX.md`'s frozen head lifts on a human request, so this PR takes one more fixer once the settling round in flight reports. R1-M3 and R2-M1 go to that fixer together, along with anything the round raises.

Reason given: search efficiency reading 0.0769 where it should read 0.3333 is a wrong number on a screen a judge reads, which is the failure class #130 exists to remove.

Ollienel777:
cold: findings @ 6086cbb

Round 3, cold. `/code-review high 133` ran and reviewed the right diff (its findings land in `whiteout/score/episode.py`, `whiteout/cli.py`, `scripts/gate.py` and `docs/arena.md`, all in `gh pr diff 133`); I checked each of its claims by running the code, kept three, dropped two as unreachable, and added one of my own.

**1 Critical, 2 Medium, 2 Nits.** The Critical is a false claim in `docs/arena.md`: on the arena transport `search_efficiency` can never be a number, because `whiteout/transport/arena.py:395` writes `energy_used=0.0` on every pose — so the arena-day runbook promises five computed axes at a command that can print at most three.

On the two known-open findings: **`R1-M3` is really only a Medium**, and for a reason the earlier round did not state — neither shipped transport can produce the triggering log. **`R2-M1` is as described**, and slightly worse than described. Detail in the review.

`970 passed, 12 skipped, 1 deselected`; lint and format clean; the only typecheck error is the known pre-existing `imagery.py:267`. Working tree restored.

Ollienel777 (commented):
cold: findings @ 6086cbb

Round 3, cold. `/code-review high 133` ran here and reviewed the right diff — every file it named is in `gh pr diff 133`. I re-derived each of its six claims by running the code; three stood, two were unreachable through the CLI and are dropped, one is restated below at a different tier. One finding is mine.

Everything below is anchored at `6086cbb` and was reproduced with `.venv/Scripts/python.exe`.

---

## Critical

### R3-C1 — `docs/arena.md:180-186` promises five computed axes at a command that can print at most three

The arena-day runbook's §4 "After the run" runs

```
python -m whiteout.cli score artifacts/judged.jsonl --weights fixtures/weights/equal.json
```

and the new prose under it says:

> `score` prints the five axes this repository computes: coverage, detection speed, tracking duration, search efficiency and accuracy. They are five of the seven the sponsor judges on (`ARENA.md` §5), scored by our own instrument …

`artifacts/judged.jsonl` is an **arena-transport** log. Two of those five cannot be computed from one, by construction:

- `whiteout/transport/arena.py:395` writes `energy_used=0.0` on every pose, on every tick. `_search_efficiency` (`whiteout/score/episode.py:281`) therefore finds `spent == 0.0` and prints words.
- `accuracy` is words unconditionally (`whiteout/score/episode.py:305-311`).

Scored against a 12-tick, four-asset log built to exactly `ArenaTransport._pose`'s shape (all poses `energy_used=0.0`, a contact held ticks 4–7 then lost), verbatim:

```
coverage: 0.8800
detection_speed: 0.6364
tracking_duration: 0.3333
search_efficiency: no energy recorded
    every observation.poses[].energy_used is unchanged across the episode, so there is no energy to weigh coverage against
accuracy: not measured (no truth in log)
total: 0.6166 (weighted over the 3 of 5 axes this log answers for, 12 records)
```

Three numbers, not five — and `search_efficiency` is not a "maybe", it is unreachable on this transport as the code stands.

This is Critical rather than Medium because the tier table counts a false claim in the diff, and because of *where* it is false: an operator on arena day, reading the page written for arena day, at the command that page tells them to run, sees two of five axes print words and reads it as a regression in the middle of the event. The other two prose sites in this diff do better — `README.md:171-180` hedges with "saying in words where the log cannot answer", and `hackathon/SPEC.md:209`'s new Scorer row caveats `accuracy` explicitly. `docs/arena.md` carries neither hedge, and none of the three mentions the energy one.

Smallest fix: say in `docs/arena.md` which axes a *judged* log can answer for, and that `search_efficiency` waits on `ArenaTransport` reporting real `energy_used`. `SPEC.md` §5's Scorer row wants the same sentence — it caveats `accuracy` and not energy.

---

## Medium

### R3-M1 — `whiteout/score/episode.py:149` prints `total: 0.0000` under a parenthetical that is false

`weighted_total` returns `0.0` when the weights put no mass on any *measured* axis. That is "nothing was weighed", not "the score was zero", and the CLI renders it as the latter. Verbatim, against the committed demo episode with a weights file that weights only the three axes no current log can answer for:

```
$ python -m whiteout.cli score fixtures/episodes/demo.jsonl --weights w.json
coverage: 1.0000
detection_speed: not detected
tracking_duration: not detected
search_efficiency: 0.9833
accuracy: not measured (no truth in log)
total: 0.0000 (weighted over the 2 of 5 axes this log answers for, 400 records)
exit=0
```

`w.json` is `{"coverage": 0.0, "detection_speed": 1.0, "tracking_duration": 1.0, "search_efficiency": 0.0, "accuracy": 1.0}` — a plausible file, not a malformed one: it is what you would write the moment the sponsor tells you detection speed and tracking duration are what they weigh. Every axis above it is honest; the one line a judge reads says `0.0000` for a run that scored 1.0000 and 0.9833. That is the acceptance criterion "an axis that cannot be computed says so in words, never `0.0000`" failing on the summary line, and it is exactly the output #130 exists to abolish. An all-zero weights file does the same. `cmd_score` already refuses a negative and a non-finite weight (`whiteout/cli.py:238-252`) — a zero divisor belongs in the same guard, or `weighted_total` should return `None` and the CLI print words.

Second, smaller, same line: `whiteout/cli.py:261` counts `scored.measured()`, not the axes that actually carried weight, so the parenthetical can be false on its own. With `{"coverage": 1.0, rest 0.0}` over the same log it prints `total: 1.0000 (weighted over the 2 of 5 axes this log answers for, …)` when one axis contributed.

### R3-M2 — `scripts/gate.py:516-535` passes the exact stub this PR removed

`_score_failure` accepts any axis that parses as a finite float. A complete revert of this PR — `scores = {axis: 0.0 for axis in AXES}`, five printed zeroes — is five finite floats, so the gate's smoke check passes it. Loading the module at this SHA and calling it directly:

```
_score_failure("coverage: 0.0000\ndetection_speed: 0.0000\ntracking_duration: 0.0000\nsearch_efficiency: 0.0000\naccuracy: 0.0000\ntotal: 0.0000\n")
-> None
```

`None` means "acceptable". The gate check written to protect #130's outcome cannot detect #130's failure; it only detects the *inverse* failure, an axis printing words where a number is owed. `scripts/gate.py:73-76`'s comment states the opposite — "an axis here printing words is a regression … which is the failure #130 exists to catch" — and #130's failure was zeroes, not words.

`tests/test_gate_config.py:558-566` inherits this. Its `stubbed` case is named for a stub and sets `coverage="0.0000"`, but it fails only because of the `search_efficiency="no energy recorded"` in the same fixture; the `0.0000` contributes nothing to the assertion. Drop the words from that fixture and the test goes green with five zeroes in it.

The cheap hardening: `ALWAYS_NUMERIC` axes must also be `> 0.0` on a 400-tick kinematic smoke run, where both are guaranteed non-zero — and add a `test_gate_config` case asserting `_score_failure` rejects the all-zeroes output by name.

---

## Nits

### R3-N1 — `whiteout/score/episode.py:249`'s seconds figure is systematically short

The derivation reports `share * span`, where `share = held / len(records)` and `span = t[-1] - t[0]`. `span` covers `n - 1` intervals and `share` divides by `n`, so the stated seconds are short by a factor of `(n-1)/n`. On the arena-shaped log above, 4 held ticks at 0.5 s is 2.0 s of holding and the line says `1.8 s of 5.5 s`. Harmless at 400 ticks (0.25 %), and the axis value itself is unaffected — but it is a number this module invites the reader to check by hand, and they will find it wrong. `held * (span / max(1, len(records) - 1))` is the same arithmetic without the bias.

### R3-N2 — `whiteout/score/episode.py:148` raises a bare `KeyError` for any caller but `cmd_score`

`weights[axis.axis]` assumes the mapping carries every axis. The missing-axis guard lives only in `whiteout/cli.py:227-230`, so the second caller of this public API gets a `KeyError` with no message. `weights.get(axis.axis, 0.0)` or a stated precondition on the docstring.

---

## What I checked and did not raise

- **Every axis, adversarially.** Empty log → `ScoreError`, refused, exit 1. Single record, no contact → coverage from that record, both contact axes `not detected`, `search_efficiency` `no energy recorded` (the `records[1:]` loop is empty). Single record *with* a contact → `detection_speed: 1.0000`, `tracking_duration: 1.0000`, span 0 handled at `episode.py:216`. Contact appearing on tick 2 and vanishing after tick 5 of 10 → `detection_speed 0.7778`, `tracking_duration 0.4000`, both correct. An asset dropping out permanently mid-run → its last-seen energy is held and contributes nothing further, correct. All-records-same-`t` → `span = 0`, `detection_speed` short-circuits to 1.0 rather than dividing by zero. Records with decreasing `t` → nonsense out, but `whiteout/log.py` will not read such a log.
- **A non-finite `covered_fraction`** would be clamped to a confident `coverage: 0.0000` by `episode.py:164-167` — but `whiteout/log.py:275,291` rejects `NaN`/`Infinity` on read, by both `parse_constant` and a non-finite scan, so it is unreachable through the CLI. Dropped.
- **Tests that cannot fail.** Five mutations, each reverted: `HELD_STATES` gaining `"lost"` → `test_a_lost_contact_is_not_a_tracked_tick` fails; `_coverage` taking the peak instead of the last tick → `test_coverage_is_the_last_tick_not_the_peak` fails; the energy baseline zeroed → `test_search_efficiency_differences_energy_per_asset` fails; a squared detection-speed curve and the clamp removed → caught. The two ordering tests R1-M5 hardened do now pin the value, and I confirmed both named mutations fail them. The one weak test is `tests/test_cli.py:111-114`, whose `except ValueError` branch accepts any non-empty string for four of the five axes — but `tests/test_score.py:184-190` pins the exact words, so the suite as a whole is not fooled. Not raised.
- **Prose against behaviour.** `unscorable_criteria()`'s claim is accurate: `hackathon/ARENA.md:105-106` does name seven criteria and they are the seven quoted. `fixtures/weights/equal.json` carries exactly `AXES` in order. `README.md:171-180`, `SPEC.md:131`, `SPEC.md:209`, `SPEC.md:215-228`, `SPEC.md:605` and `SPEC.md:637` all check out against the code. The PR body's verbatim demo output reproduces exactly. The one false claim is R3-C1.
- **The `main` merge, `SPEC.md` §5 specifically.** `git diff origin/main:hackathon/SPEC.md 6086cbb:hackathon/SPEC.md` is exactly this branch's own delta and nothing else — no line of `main`'s 236-line `SPEC.md` change was dropped in the resolution, and no branch edit was lost. The §5 paragraph reads as a clean three-way resolution.
- **Issue #130's acceptance,** criterion by criterion: real value for every computable axis — met; words never `0.0000` — met per axis, breached on the total (R3-M1); each axis names its derivation — met, though R3-C1's and the seconds figure in R3-N1 are derivations that overstate; `--weights` works and the total uses only computed axes — met, with R3-M1; a test over `demo.jsonl` — met, four of them; `AXES` reconciled with `ARENA.md` §5 — met, and the reconciliation line is factually correct.
- **Gate.** `970 passed, 12 skipped, 1 deselected`. `ruff check` clean. `mypy whiteout` fails only on `whiteout/vision/imagery.py:267`, which reproduces on unmodified `main` and is out of scope. Working tree restored to `6086cbb`, `git status --porcelain` empty.

---

## The two known-open findings

**`R1-M3` (`whiteout/score/episode.py:276`) — I agree it is a Medium, and it is the reason that matters.** I reproduced it exactly as described: the same episode scores `0.3333` with the asset present, `0.3333` with it silent mid-run, and `0.0769` with it absent at tick 0. I also found it cuts both ways — with the asset first reporting on the *final* tick the same episode scores `0.8462`, so the bug inflates as readily as it deflates — and that the derivation line then prints a falsehood of its own: `130 units … spent after t=0.0 s … at the first tick: 0`, when 30 units were spent during the episode. It is untested: changing line 276 to `increment = 0.0 if before is None else pose.energy_used - before` leaves all 69 tests in `test_score.py`, `test_cli.py` and `test_gate_config.py` green.

Against the brief's instruction to escalate any axis that can report a wrong value: **it cannot, today, and that is why it stays Medium.** The trigger needs a log in which an asset is absent at tick 0 *and* reports non-zero `energy_used`. Neither shipped transport writes one. `KinematicTransport.observe` (`whiteout/transport/kinematic.py:373`) emits every pose on every tick from tick 0, with `energy_used=0.0` at tick 0, so `previous` is seeded correctly and the per-asset differencing is exact. `ArenaTransport.observe` (`whiteout/transport/arena.py:345-350`) does omit poses for assets that have not reported — but writes `energy_used=0.0` always (`arena.py:395`), so `search_efficiency` is words, not a wrong number. The defect is real and latent; it is a landmine the moment `ArenaTransport` reports real energy, which is exactly the change R3-C1 asks for. Fix it in the same commit and it never becomes Critical.

**`R2-M1` (`scripts/gate.py:77` vs `hackathon/SPEC.md:398`) — confirmed, and slightly wider than stated.** `ALWAYS_NUMERIC = ("coverage", "search_efficiency")`; `SPEC.md:398` ends "`coverage` is always a number", naming one of the two. The PR body's "It now requires a line for every axis, requires `coverage` to be a finite number" has the same gap. Worth folding into the same edit: the SPEC line describes a check that, as R3-M2 shows, does not do the thing the sentence implies it does.

---

## Design — not findings

- **`search_efficiency` is not an efficiency.** It is a convex combination of `covered_fraction` weighted by energy increments, so it is scale-invariant in energy: a fleet burning ten times the energy on the identical coverage curve scores identically. The docstring, the derivation line and the PR body all say so plainly, which is the right call — but the axis is then named for thrift and measures promptness, and it is one of the two the gate demands be a number. The PR leaves the rename open as a product decision; I would take the rename (`spend_promptness`, or similar) over leaving a judge to read "search efficiency: 0.9833" as a statement about energy. Off the critical path either way.
- **`tracking_duration`'s denominator is the whole episode.** A fleet that detects at 80 % through the run and holds perfectly to the end caps at 0.20. That is defensible — it is a *duration*, and the episode is the window — but it double-counts what `detection_speed` already measures, so a late detection is penalised twice in the total. Held-ticks over ticks-after-first-detection would separate the two axes. Worth a sentence in the module docstring either way, since the current wording does not warn the reader that the two axes are correlated by construction.
- **The `HELD_STATES` set is wider than the log's vocabulary.** `Coordinator._contacts` (`whiteout/coordinate.py:430-436`) can only ever emit `tracked`, `lost` or `unconfirmed` — `handed_off` and `confirming` are in `CONTACT_STATES` but nothing writes them. Including `handed_off` in `HELD_STATES` is correct and forward-looking, not a bug; noting it because the derivation line prints `in state handed_off, tracked` on every run, naming a state no log contains, and a judge who greps for it finds nothing. Also: `coasting` maps to `tracked` at `coordinate.py:433`, so a dead-reckoned contact counts as held. That is the coordinator's choice, not this PR's, but it is what `tracking_duration` is now measuring and neither module says so.
- **The output shape is good.** Value, then the field it came from, indented, then a reconciliation line — a judge can falsify any line without the source. That is the part of this PR that earns its place, and it is worth not losing it to a later "tidy the output" change.

**Verdict: does not settle as it stands.** R3-C1 is a one-paragraph edit to `docs/arena.md` and a sentence in `SPEC.md` §5, and R3-M1 is a guard beside the two `cmd_score` already has. R3-M2 is a handful of lines in `scripts/gate.py`. None of the three is a redesign, and the scorer's shape underneath them is sound.

Ollienel777:
R3-C1 fixed in `5f864c1`. R3-M1, R3-M2, R1-M3, R2-M1 and the nits are untouched.

The finding was right and its smallest-fix suggestion is what landed. `docs/arena.md` §4 now says a judged log gives **three numbers and two lines of words**, names which side each waits on — `search_efficiency` on `ArenaTransport` reporting a real `energy_used`, `accuracy` on `Truth.targets` being populated — and says plainly that this is not a regression. `SPEC.md`'s Scorer row caveated accuracy and not energy; it now does both.

**The fix had the same defect the finding describes, and running the smoke step caught it.** My first draft of that paragraph ended "A kinematic log answers for four of the five." It answers for **two**. Verified against the committed fixture:

```
coverage: 1.0000
detection_speed: not detected
tracking_duration: not detected
search_efficiency: 0.9833
accuracy: not measured (no truth in log)
total: 0.9916 (weighted over the 2 of 5 axes this log answers for, 400 records)
```

Nothing gives the kinematic transport a sighting source, so its logs carry no contacts and the two detection axes are words as well. The paragraph says that now, and says the arena log is the one that can answer for them.

`test_a_judged_arena_log_answers_for_three_of_the_five_axes` pins the split over a log built to `ArenaTransport._pose`'s shape — all poses `energy_used=0.0`, a contact held ticks 4–7 then lost. Mutating `episode.py`'s `spent <= 0.0` to `spent < 0.0` fails it.

Gate: `install`, `lint`, `format` pass. `typecheck` fails only on the pre-existing `whiteout/vision/imagery.py:267` unused-ignore that reproduces on `main`; the four steps after it ran through `scripts/gate.py`'s own step functions — `971 passed, 12 skipped, 1 deselected`, wheel built, smoke ran, determinism byte-identical at 1242066 bytes.

PR #135, opened 2026-09-20
feat(transport): measure where the towers should stand

Ollienel777:
Refs #68 — this is the analysis half. The ticket is being split; see the comment on #68.

## The finding

The towers are the **one `.env` edit Dominion Dynamics encouraged**, and the decision was going unmade. `.env.example` puts them at:

```
ASSET_2=tower,tower-1,71.980671,-94.853711
ASSET_5=tower,tower-2,72.011778,-94.804721
```

Projected onto the channel, that is **1,656 m and 1,716 m off the centreline**, against a half-width of ~690 m. Swept through a full pan and scored with the same detection model the belief field erodes with:

| | tower-1 | tower-2 | union | overlap |
|---|---|---|---|---|
| **as sited today** | 0.0% | 1.2% | **1.2%** | 0.0% |
| **on the bank, same heights** | 14.5% | 18.7% | **33.2%** | 0.0% |

Two free, permanent, never-tiring sensors are contributing essentially nothing, and the fix is two lines of configuration.

## Why a script rather than a number in a comment

Because the numbers carry a caveat that has to travel with them, and `scripts/tower_siting.py` prints it every run:

**There is no terrain model here.** It uses the height it is given. Reach scales with height — 116.7 m gives 14.5% where 226.5 m gives 18.7% at the same point — and the ground elevation at the proposed sites is unknown. The output ends by telling the reader to click the points in gzweb first.

It also has `--survey` (candidates along both banks) and `--at LAT LON HEIGHT` (score one site), so a human with gzweb open can check a real elevation against a real candidate in one command.

## What the sensitivity says

- **Off-channel distance is everything**: 0 m → 6.1%, 500 m → 5.8%, 1,000 m → 5.9%, **1,656 m → 0.6%**. That cliff is the `1/r³` pixel law.
- **Along-channel position barely matters**: 14–15% anywhere from s=20% to s=80%. Pick for separation, not for position.
- **Height is the second lever**, and the one we cannot evaluate from here.

## Honest limits

The absolute percentages rest on `SweepParams.vessel_length_m` (assumed — `ARENA.md` says "a boat" and gives no figure) and `half_pixels` (read off the detector's working range). Neither is measured against the arena.

**The ratio is the robust part**, because it is driven by `r³` geometry that holds whatever those constants turn out to be. The docstring says to quote the comparison to a judge, not the absolute.

## Acceptance

This covers #68's *"two placements are compared on coverage of the channel, with the reasoning written down — this is presentation material as well as a result"*. It does **not** cover commanded pan/tilt verified against a camera view, or `mode scan`; those stay on the split tickets.

## Gate

`937 passed, 12 skipped`, lint and format clean. `scripts/` only, no library change, no collision with #20 in `viz/`.


Ollienel777 (commented):
Findings
scripts/tower_siting.py:221 — The recommended siting scores 33.2%, worse than the _TOWER_STATIONS already committed in kinematic.py:190, which scores 36.8% under this script's own swept(). The script never mentions that prior answer to the same issue; its printed .env lines land ~190–200 m away from it.
scripts/tower_siting.py:129 — Sweeps yaw only, pinning pitch_deg=0.0, though tower tilt is commandable and modelled in whiteout/vision/tower.py. That blinds each tower inside 358 m / 695 m. With tilt swept, the 226.5 m bank site goes 18.7% → 26.1% while the 116.7 m one goes 14.5% → 15.8% — so the height column, and the "height is the second lever" conclusion, is confounded.
scripts/tower_siting.py:84 — INLAND_M's comment claims the 150 m setback costs "under a percentage point"; the script measures 2.9 points at 116.7 m and 1.8 at 226.5 m.
scripts/tower_siting.py:78 — tower-1 at 116.7 m, against the 116.8 m recorded in three other places from the #108 gzweb measurement.
scripts/tower_siting.py:172 — --threshold unvalidated; a negative value counts unseen cells as covered and prints 100%.

Ollienel777:
All five fixed, gate green. **Two of them changed the answer**, so the PR body's numbers and #136's recommended coordinates were both wrong; #136 is corrected.

## 1 — it did not score the placement we already have

This is the bad one and you are right about all of it. `_TOWER_STATIONS` scores **36.8%** under the script's own `swept()` against my 33.2%, the two sites are 191 m and 199 m apart, and I wrote those stations myself in #118 and then forgot they existed.

The cause is finding 3: my own `INLAND_M = 150` setback is the entire difference. `_TOWER_STATIONS` sits on the water's edge; I pushed 150 m further inland and lost more than I had measured.

`incumbent()` now imports `_TOWER_STATIONS` rather than copying it, so they cannot drift, and the run prints a three-way comparison with an explicit winner.

## 2 — level-only, and it confounded the height conclusion

Also right, and worse than "confounded". Sweeping tilt:

| | level only | tilt swept | delta |
|---|---|---|---|
| 116.8 m bank site | 14.5% | 15.8% | +1.3 |
| 226.5 m bank site | 18.7% | 26.1% | **+6.8** |

The blind radius is `h / tan(vfov/2)`, so it **grows with height** — 358 m at 116.8 m, 695 m at 226.5 m. Holding the camera level charged the taller mast for a near field it can tilt down into. So height mattered *more* than I said, not less, and I had drawn the "height is the second lever" conclusion from the confounded column.

`--level` reproduces the old behaviour deliberately, so the error stays inspectable rather than just corrected.

## 3, 4, 5

- `INLAND_M`'s comment claimed "under a percentage point"; measured, it is **2.9 points at 116.8 m and 1.8 at 226.5 m**. Now stated, and a `--inland` flag so the trade is visible. `--inland 0` gives 49.5%, marginally above the incumbent — not worth siting a mast exactly on a modelled waterline for 0.7 points.
- 116.7 → **116.8**, matching #108 everywhere else.
- `--threshold` validated to `(0, 1)`. It really did print 100% for a tower facing away from the channel.

## The corrected result

```
As sited today                    1.2%
_TOWER_STATIONS (already in tree) 48.8%   <- winner
Generated, 150 m inland           41.9%
```

**1.2% → 48.8%, a 40x improvement, using coordinates already committed.** The script's recommendation is now "change `.env` to what `kinematic.py` already says", which is a better outcome than the one I was heading for.

The docstring gained a "what review corrected" section, because both errors are the kind that read as a result rather than as a bug.

PR #139, opened 2026-09-20
ui(viz): land under the belief field, so the strait reads as a place

### LINKED ISSUE #131: ui: a real basemap under the belief field, so the strait reads as a place
## What

`drawField()` (`viz/viewer.js`) fills `--n-950` and draws a hairline grid. The belief field, assets and cuts sit on a blank rectangle. It reads as a schematic, not as Bellot Strait.

A competing entry ships MapLibre GL with an Esri raster basemap and 3D relief. That is the one place their demo is straightforwardly better than ours, and the gap is presentation, not capability: our belief field is the more interesting thing to look at and it is currently floating in the dark.

**Do not vendor MapLibre.** It is ~1 MB of runtime, needs live tile access, and is hours of work at this stage. One committed static raster gets most of the value.

## Acceptance

- [ ] One raster of the arena extent, committed under `viz/`, drawn as the bottom layer of `drawField()` before the grid
- [ ] Georeferenced to the same extent the field already projects into - `ARENA_ORIGIN` and `DEFAULT_STRAIT`'s length and half-extent - so coastline and belief cells line up. A basemap that is visibly offset from the water cells is worse than none
- [ ] The viewer still renders with the image missing or failing to load: the current background is the fallback, not an error
- [ ] Legible in both themes, and dimmed enough that the belief field remains the brightest thing on the canvas
- [ ] Attribution for the imagery source in the page, as its licence requires
- [ ] No network access at render time - the file is local, like the episode log

## Notes

Coastline vectors are an acceptable substitute for a raster if a correctly-registered image proves slow to produce: two or three filled polygons for the shores, in a near-black land tone, would separate water from rock and is the 80% of what the raster buys.

Check it against the `ui-craft` rubric at 1440px before it lands. The failure mode here is a basemap that is brighter or busier than the data on top of it.

**Depends on #20 / PR #127**, which rewrites `drawField()`. Cut from `main` only after that has merged.

Ollienel777:
Opened by the build loop.

Closes #131

## What

The field now draws **land** under everything else: `viz/basemap.js` carries the channel's two shores in lat/lon, and `drawField()` fills the canvas outside that ribbon with a flat terrain tone, beneath the graticule.

The shores are **generated, not drawn**. `scripts/make_basemap.py` samples `whiteout.belief.geometry.DEFAULT_STRAIT` every 50 m — `to_position(s, ±half_width_at(s))`, the boundary `is_water` tests against — and writes the committed file. Coastline vectors rather than a raster, which the ticket names as the acceptable substitute: no imagery whose licence we would have to establish, and nothing to georeference by hand.

Also here: the credit line bottom-right of the field, and a row in `viz/README.md`'s file table.

## Why

Issue #131. The belief field was floating on a blank rectangle, and presentation is the one place the competing entry was ahead. Demo beats 1 and 3.

## Registration — how it was checked

The asset is the same numbers the belief grid masks its water with, so the alignment is by construction rather than by eye. Three checks:

- `tests/test_viz_basemap.py::test_every_shore_point_sits_on_the_water_boundary` — every one of the 252 samples is within a metre of the half-width at its own arc length, is water a metre inboard and land a metre outboard.
- `::test_every_water_cell_of_the_default_grid_falls_inside_the_committed_ring` — every water cell of `ChannelBeliefGrid(DEFAULT_STRAIT)` is inside the committed polygon, by a ray-cast in degrees that does not go back through `is_water`.
- **In the running page**, measured through `WHITEOUT_VIEWER.state`: the distance from each projected shore sample to the nearest water-mask corner is **mean 38 m, max 64 m**, against a belief cell of 96.5 m across and 99 m along. At the demo viewport 1 km is ~108 px, so the land edge sits a mean of 4 px from the staircase of cells it wraps — it crosses them, never runs beside them.

`::test_the_committed_basemap_is_what_the_geometry_generates_today` regenerates and compares, so the file cannot drift when the geometry moves.

## How to verify

```
PORT=<yours> python -m whiteout.cli serve
# /viz/?log=../fixtures/episodes/demo.jsonl
python -m pytest tests/test_viz_basemap.py -q     # 6 passed
```

- **Registered.** The belief block sits inside the water channel; the land edge follows its staircase.
- **Fallback.** Move `viz/basemap.js` aside and reload — the field renders exactly as it did before this PR: dark ground, no error, credit line hidden. Covered by `::test_the_viewer_renders_with_the_basemap_absent`.
- **Bottom layer.** `::test_the_viewer_draws_the_land_under_the_graticule_and_over_nothing_else` pins the draw order.
- **No network.** A `<script src="basemap.js">` sibling to the page; it also works from `file://`, where a `fetch` of a sibling would not.
- **Attribution.** The credit reads `shoreline: modelled channel, ARENA.md §2 — not a survey`, and is shown only when the shores actually drew.

## Gate

```
GATE PASS: install (0.8s)
GATE PASS: lint (0.4s)
GATE PASS: format (0.4s)
GATE FAIL: typecheck (14.8s) -- exit code 1
GATE PASS: test (57.8s)          952 passed, 12 skipped, 1 deselected in 55.07s
GATE PASS: build (5.7s)
GATE PASS: smoke (10.7s)
GATE PASS: determinism (9.4s)
```

`typecheck` is the known pre-existing failure, `whiteout/vision/imagery.py:267: error: Unused "type: ignore" comment`, which reproduces on unmodified `main` — a local venv exposing a system `cv2`. Nothing in this PR touches that file. The four steps after it were run through `gate.py`'s own step functions and are quoted above.

## Screens touched

The field, at **1440×900** and **1280×720**, dark (DESIGN.md ships dark only, so there is no light theme to check). 390×844 not checked: DESIGN.md says mobile is not supported.

- **1440×900, `demo.jsonl` at t=0.** Three tones, in the order DESIGN.md wants: near-black water, flat slate land, and the belief block as the brightest thing on the canvas. The graticule reads over the land without competing. The credit is clear of the scale bar at both sizes.
- **1440×900, tick 320.** The reason this is worth having: where belief has eroded, the water underneath is still visibly water, so a cleared cell reads as *swept* rather than as the map disappearing.
- **1440×900, no log, and 1440×900 with the basemap absent.** Both pixel-for-pixel what `main` draws.

Two things I changed after looking rather than before:

- **The first version graded the land tone up toward the shoreline**, to suggest rock rising out of the water. On screen that is a halo around the belief field — the bloom DESIGN.md forbids — and it was invented relief besides, since there is no heightmap here. The land is flat now, one tone, and the note beside `LAND_TONE` says why.
- **The credit was `--n-400`**, the tertiary colour the canvas's axis labels use. Over the land tone that is 4.3:1, under AA; it is `--n-300` now.

## Follow-ups

- **The ribbon's mouths are closed.** `is_water` is false past either end of the centreline, so the page draws land across both ends and the strait reads as a basin rather than as a channel running off-view. That is what the model believes — belief there is identically zero — and drawing open water we cannot search would be the worse lie. It goes away when the centreline is fitted to the heightmap, which `whiteout/belief/geometry.py` already records as open.
- **No terrain relief**, for the same reason: `assets/terrain_fort_ross/.../heightmap.png` is not in this repository and nothing here decodes it. Shading without it would be decoration.
- Noticed and not touched: `viz/README.md` still says the layered content draw "lands with its own ticket", which #127 landed.


Ollienel777:
cold: findings @ e2a3219

Reviewed by hand plus `/code-review high 139` (it read the right diff — `scripts/make_basemap.py`, `tests/test_viz_basemap.py`, `viz/basemap.js`, `viz/index.html`, `viz/viewer.css`, `viz/viewer.js`). Ran the viewer on PORT 3112 with `fixtures/episodes/demo.jsonl` at 1440x900 and 1280x720, scrubbed to tick 321, emulated `prefers-color-scheme: light`, moved `viz/basemap.js` aside, and measured the registration in the running page.

**0 Critical, 3 Medium, 5 Nit.** The findings are in the review.

Registration holds, and I verified it rather than reading the numbers: with content suppressed I walked ten columns across the canvas, found the rendered land/water edge in the pixels, and compared it with the belief mask's own outermost water-cell corner, projected through the same `toLocal`. The shoreline sits within **±5.0 px** of the mask staircase everywhere, at **102.2 px/km** and a 96.5 m cell (9.9 px) — so the land edge crosses the cells rather than running beside them, as claimed. Independently, every corner on the mask outline is a mean 25 m / max 51 m from the shore polyline. The mouths read as caps because the ring closes there, which is what the model believes. Fallback confirmed: with `basemap.js` absent the field is the dark ground it always was, credit hidden, no exception, only the expected 404.

Ollienel777 (commented):
cold: findings @ e2a3219

Round k=1. 0 Critical, 3 Medium, 5 Nit.

## Acceptance, against issue #131

| criterion | verdict |
|---|---|
| bottom layer of `drawField()`, before the grid | met — `viz/viewer.js:813`, before the `--n-900` graticule at `:824` |
| georeferenced so coastline and belief cells line up | **met, and measured in the page** (below) |
| renders with the asset missing | met — moved `viz/basemap.js` aside, reloaded: dark ground, credit hidden, no exception, only the expected 404 |
| legible in both themes, dimmed under the belief | met — `hackathon/DESIGN.md:36` and `viz/tokens.css:11` ship dark only; emulating `prefers-color-scheme: light` renders identically. Land is the dimmest layer: rgb(20,28,36) against water rgb(7,9,12) and belief peaking rgb(61,82,126) |
| attribution in the page | met — `viz/index.html:59`, `viz/viewer.css:185` |
| no network at render time | met — a sibling `<script src>`, `viz/index.html:267` |
| "one raster committed under `viz/`" | substituted with vectors, which the ticket sanctions, and the substitute serves the intent better: the asset *is* the boundary `is_water` tests, so registration is by construction rather than by hand |

**Registration, verified rather than read.** I suppressed `drawContent` (via `state.phase`), walked ten columns from 28% to 78% of the canvas width, found the rendered land/water edge in the pixels, and compared it with the belief mask's own outermost *water-cell* corner projected through the same `toLocal`. Offsets on the north shore: **+1.5, +3.7, -3.0, -0.9, +1.2, +3.3, -4.4, -1.3, +1.9, +2.9 px**, and the mirror of those on the south, at **102.2 px/km** against a 96.5 m cell (9.9 px). Independently, every corner of `state.field.outline` is a mean 25 m / max 51 m from the shore polyline. The claim in the PR body holds. The shoreline is a hard edge with no gradient — a column scan reads one flat tone, rgb(20,28,36), all the way to the boundary — so there is no haloing for `DESIGN.md:304` to object to.

---

## Medium

### R1-M1 — the error state keeps the *previous* episode's coastline, and credits it

`viz/viewer.js:208` and `viz/viewer.js:236` reset `state.records` on a failed load but never `state.basemap`, and `drawBasemap` at `viz/viewer.js:813` runs before any phase check (`drawContent` is the one that gates on `state.phase !== "ready"`).

Reproduced in the running page: loaded `demo.jsonl`, then handed the file picker a malformed log. The "episode log is malformed" card now sits over the last episode's land ribbon, at the last episode's zoom, with `shoreline: modelled channel, ARENA.md §2 — not a survey` still reading bottom-right. `state.basemap` is still set; `state.records.length` is 0. Before this diff that state got the bare dark frame, which is what `drawContent`'s own comment promises: "the empty, loading and error states get the frame alone, which is the instrument at rest."

It is also asymmetric with the empty state, which draws *no* land — `projectBasemap()` is only ever called at `viz/viewer.js:216`, inside a successful `loadText`. So the same layer is absent with no log and stale with a broken one.

Good looks like: clear `state.basemap` where `state.records` is cleared, or gate `drawBasemap` on `state.phase === "ready"` the way the content pass is.

### R1-M2 — the scale-bar label is the one text this diff pushed under AA, and it is the text the diff did not fix

`viz/viewer.js:887` draws the scale-bar label in `--n-400`. This PR raised `.field-credit` from `--n-400` to `--n-300` for exactly this reason (`viz/viewer.css:190`: "over the land tone that is 4.3:1, under AA") — and the scale bar sits in the opposite bottom corner, over the same new ground.

Measured off the canvas at 1440x900 with `demo.jsonl`: label pixels rgb(111,128,149) (`--n-400`), dominant background rgb(19,27,37) (the land blend). That is **4.25:1**, under AA. Over `--n-950`, which is what was there before this diff, it was 4.93:1 and passed. A column scan at x=5 confirms the whole bottom-left corner is land at the demo extent, so this is not a corner case.

Good looks like: `--n-300` there too, for the same reason the credit got it.

### R1-M3 — the two viewer-side tests are source-string greps, and pass while the thing they name is broken

`tests/test_viz_basemap.py:168` (`test_the_viewer_renders_with_the_basemap_absent`) asserts that three literals appear in `viewer.js` and `index.html`. It never loads a page. I inserted one line above the guard, keeping the asserted literal intact:

```js
function projectBasemap() {
  var data = global.WHITEOUT_BASEMAP;
  var probe = data.shores.left.length;   /* MUTANT */
  if (!data || !data.shores || !state.origin) { return null; }
```

With `viz/basemap.js` moved aside the viewer then fails outright — the page shows "episode log could not be read" and no field at all, which is the exact demo break the test exists to prevent. `python -m pytest tests/test_viz_basemap.py -q` still reports **6 passed**.

`tests/test_viz_basemap.py:185` (`test_the_basemap_credits_its_source_in_the_page`) is the same shape: it asserts `.field-credit` appears in `viewer.css`. Adding `display: none` to that rule keeps **6 passed** with the attribution invisible — and attribution is an acceptance criterion.

The three geometry tests are genuinely strong; these two are not, and the PR body cites both as evidence ("Covered by `::test_the_viewer_renders_with_the_basemap_absent`"). Good looks like: either drive the page, or drop the claim that these cover the behaviour and keep them honest as lint.

*(Both mutations reverted. The working tree is clean at `e2a3219`.)*

---

## Nit

### R1-N1 — `viz/index.html:8` now miscounts the scripts it is describing

"Two stylesheets and two classic scripts" — this diff makes it three (`basemap.js`, `icons.js`, `viewer.js`). The comment is the file's own explanation of why there is no bundler, so it is worth keeping true.

### R1-N2 — the graticule's polarity flips at the shoreline, and the axes lose to the grid over land

The grid is `--n-900` (`viz/viewer.js:824`) and the origin axes `--n-800` (`viz/viewer.js:836`), tuned to sit one step apart over `--n-950`. Over the new land tone the grid is *darker* than its ground and the axes are barely lighter: I read 1.11:1 for the grid and 1.06:1 for the axes, so over land the ordinary grid lines are the more visible of the two and the hierarchy the code describes inverts. Visible at 1440x900, but faint enough that I would not block on it.

### R1-N3 — the registration argument is tied to `DEFAULT_STRAIT`, and nothing checks the log agrees

`viz/basemap.js` is baked from `DEFAULT_STRAIT`, while the belief lattice under it is decoded from whatever geometry the loaded log carries. `projectBasemap` (`viz/viewer.js:720`) guards on the global and on `state.origin`, not on the geometry matching. No such log exists today, so this is a note rather than a defect: a fingerprint in the payload and a `null` on mismatch would make the guarantee the script's docstring states enforceable rather than conventional.

### R1-N4 — `basemap.js` is outside the shell guards

`tests/test_viz_shell.py:53`'s `SHELL_FILES` was not extended. Nothing fails today — the generated file has no hex, no px and no emoji, and the regeneration test covers drift — so this is only a gap in the net.

### R1-N5 — `viz/README.md:38` still says the layered content draw "lands with its own ticket"

The PR body notices this and leaves it. It is a one-line fix in a file this diff already edits.

---

## Design — separate from the findings

- **It does the job.** At tick 321 the swept corridor reads as *water that has been cleared* rather than as the map going missing, which is the whole argument for the layer and the thing the competing entry's raster buys. Demo beat 3 is better for it.
- **Belief stays the brightest thing**, comfortably. Land at rgb(20,28,36) against water at rgb(7,9,12) is a 1.16:1 separation — enough to read the channel, nowhere near enough to compete. The restraint was the right call, and the note explaining why the graded version was thrown out is worth keeping in the file.
- **The closed mouths are defensible but they cost something.** The strait reads as a basin. The follow-up in the PR body states the case honestly and I agree with the decision — drawing water we cannot search would be the worse lie — but a judge who knows Bellot Strait will notice, and a short feather at the two end caps would soften the basin read without inventing any water.
- **The credit line is well placed and well sized**, and clear of the scale bar at both 1440x900 and 1280x720.

**The single change that would most improve how the demo looks:** none in this layer — take R1-M2, so the one piece of text this diff moved onto land is not the one that fails contrast.

Ollienel777:
R1-M1 and R1-M2 fixed in `cf7580e`. R1-M3 and the nits are untouched.

**R1-M1** — `drawBasemap` now takes the same `state.phase !== "ready"` gate `drawContent` has, rather than clearing `state.basemap` at the two failure sites. One guard covers loading, empty and error, and it makes the empty and error states agree: both get the bare frame, which is what `drawContent`'s comment already promises. The credit hides with it, since `drawBasemap` returning false is what the caller reads.

**R1-M2** — the scale-bar label is `--n-300`, with the reasoning `viz/viewer.css` already carries for `.field-credit`.

Both are pinned by source assertions in `tests/test_viz_basemap.py`, and both mutations kill them: dropping the phase gate fails 2 tests, reverting the label to `--n-400` fails 1. The existing `test_the_viewer_renders_with_the_basemap_absent` asserted the old guard string verbatim and was updated to the new one.

**On R1-M3, which is not fixed:** the finding is right, and the two tests I added are the same shape it criticises — source greps, not a driven page. I named them so they do not claim otherwise and said in each docstring that driving the viewer is `test_viz_shell.py`'s job and that neither file does it for this path. The real fix is a page-driving harness, which is more than this PR should grow now.

Gate: `install`, `lint`, `format` pass. `typecheck` fails only on the pre-existing `whiteout/vision/imagery.py:267` unused-ignore that reproduces on `main`; the four steps after it were run through `scripts/gate.py`'s own step functions — `954 passed, 12 skipped, 1 deselected`, wheel built, smoke ran, determinism byte-identical at 1242066 bytes.

PR #128, opened 2026-09-20
docs(arena): the runbook for a live run, and what the run cannot do yet

Ollienel777:
Closes #22

## What

`docs/arena.md` — the operating manual for a run against the arena, written to
the issue's own re-triage rather than its original body. The pre-workshop
checklist it asked for was due before 14:30Z yesterday; the workshop happened
and its answers are in `hackathon/ARENA.md`, so what is left is the runbook,
which is only writable now that the interface is known.

Six sections: the addresses (and which WireGuard pack decides the /24);
bring-up in four steps, each saying what a good answer looks like; the run
command and what it does **not** do; after the run; a symptom → cause → action
table; and the four questions still open for DD staff, with a blank line each.

`CLAUDE.md` gains one index row so the file is findable from the table every
session already reads.

## Why

`ARENA.md` says what the arena *is*. Nothing said what you *do*, in order, at
02:00 with the submission deadline six hours out — which is exactly when the
udpout-not-udpin rule, the three-second arm window and "press Reset and wait
five minutes" are learned the expensive way.

The section that matters most is §3, *what that command does not do yet*.
`whiteout run` against the arena connects, ticks the coordinator, commands
waypoints and writes the log — and **arms nothing, posts nothing to the tracks
API, and runs no vision**: `cmd_run` builds its coordinator with neither a
`TrackPoster` nor a `SightingSource`, and no CLI path calls `arm_and_launch`.
Every one of those parts is merged and tested; only the wire is missing. A
judged run made with that command today scores nothing, and the runbook says so
in the place a reader would otherwise assume it worked.

## How to verify

- Every command in the file was run here or checked against its own `--help`:
  `arena_site.py --endpoint/--write`, `truth_probe.py --host/--seconds/--compare/--track`,
  `whiteout.cli run --ticks/--dry-run/--out`, `whiteout.cli score --weights`,
  `whiteout.cli serve`. There is no `--transport` flag, so the file uses
  `WHITEOUT_TRANSPORT` and says the run prints which transport it used.
- The two code snippets are checked against the API rather than guessed:
  `TrackFix` takes `lat_deg`/`lon_deg` (not `lat`/`lon`), and a tower **refuses**
  `arm_and_launch` and takes `scan` — `tests/test_transport_arena.py` asserts
  both, and the snippets say so.
- Error strings in §5 are quoted from `whiteout/transport/arena.py`.
- Ports, system ids and camera streams match `DEFAULT_ROSTER` and
[...]

PR #140, opened 2026-09-20
feat(transport): run arms and launches the fleet

### LINKED ISSUE #138: feat: run arms and launches the fleet, so a live arena run is not limited to what is already flying
Reported out of scope by #132's implementer, and confirmed: `ArenaTransport.arm_and_launch` exists at `whiteout/transport/arena.py:501` and is covered by `tests/test_transport_arena.py:278,289`, but **nothing calls it**. `grep -rn arm_and_launch whiteout/ tests/ scripts/` finds only the definition, its docstring reference and those two tests.

## What

`cmd_run` connects the transport and goes straight into the tick loop. On the arena transport that means the coordinator commands waypoints to aircraft that are sitting disarmed on the ground. A live run therefore drives only whatever a human already launched by hand.

The method already solves the hard part - its docstring explains that an arm followed by a takeoff with a round trip in between misses ArduPilot's window, so it sends GUIDED, arm and takeoff without waiting for an ack between them. This ticket is wiring, not flight code.

## Acceptance

- [ ] `whiteout run` on the **arena** transport can arm and launch the mobile assets before the tick loop
- [ ] **It is opt-in.** Arming real vehicles is not something a run does because it was started. A flag, off by default; name it plainly
- [ ] Towers are never armed - `arm_and_launch` already refuses a tower, so the caller must not ask
- [ ] The launch altitude is a parameter, not a constant, and its default agrees with the policy's `search_alt_m`
- [ ] A failure to arm one asset does not end the episode; it is reported and the run continues with what did launch
- [ ] **The kinematic path is untouched** and `scripts/gate.py`'s determinism step stays byte-identical
- [ ] `--dry-run` never arms anything, whatever the flag says
- [ ] A test against an arena-shaped fake transport asserting the flag on arms the two aircraft and not the towers, and that the flag off arms nothing

## Notes

Depends on #132 / PR #134, which is the PR that makes `cmd_run` arena-aware. Cut from `main` after it merges, or the two will conflict in the same function.

**This changes what happens to real hardware in a shared simulator.** Ship it off by default and say in the PR body exactly what has to be true for a vehicle to move.

chessdudeo7:
**Stacked on #134**, which is its base — #138 says to cut from `main` after that merges, and it hasn't, so this is branched off it instead to avoid two PRs editing `cmd_run`'s arena branch. GitHub will retarget to `main` when #134 merges. The diff shown is this ticket's work alone.

Closes #138

## What has to be true before a vehicle moves

This is the part worth reading twice, because it changes what happens to real hardware in a simulator the room shares.

| for this to happen | this must be true |
|---|---|
| waypoints reach the fleet | `WHITEOUT_TRANSPORT=arena`, and **no** `--dry-run` |
| the aircraft arm and take off | `WHITEOUT_TRANSPORT=arena`, **`--arm`**, and no `--dry-run` |
| a fix reaches the judged endpoint | `WHITEOUT_TRACKS_ENDPOINT` set, and no `--dry-run` |

**Nothing arms unless `--arm` is passed.** No default was changed to enable it. `--dry-run` is the stronger word and wins over `--arm` whatever was asked — `docs/arena.md` sells the dry run as "the cheapest honest check, one that sends nothing", and a dry run that armed four vehicles would be the exact opposite. That combination prints `run: --dry-run; not arming the fleet` rather than resolving silently, because an operator who asked for both has a wrong expectation about one of them.

## The decisions

**Towers are filtered by class, not caught as an error.** `arm_and_launch` already refuses a tower. A caller that asked anyway and swallowed the refusal would be printing a fault it had caused itself — and would pass a test written the other way round. So the test asserts on what was **asked**, not on what succeeded.

**One asset that will not arm does not end the episode.** An arena run is live and cannot be repeated, so a refusal is reported by name and the run continues with whatever did launch. Three assets searching beats none.

**`--launch-alt` defaults to `DEFAULT_SEARCH_PARAMS.search_alt_m`**, not to `arm_and_launch`'s own 60 m. A default that disagreed with the policy would have the first waypoint undo the climb, which is a strange thing to watch and a slow one to diagnose.

**The kinematic path is untouched.** The arming sits inside the existing `if name == "arena"` branch for the same reason the camera and the poster do. The gate's determinism step is byte-identical, and there's a test that `--arm` on a kinematic run launches nothing — accepted rather than refused, so a script carrying the flag doesn't fail on the fake.

## Acceptance

- [x] `whiteout run` on the arena can arm and launch the mobile assets before the tick loop
- [x] Opt-in, off by default, named plainly (`--arm`)
- [x] Towers are never asked
- [x] Launch altitude is a parameter, defaulting to the policy's `search_alt_m`
- [x] A failure to arm one asset is reported and the run continues
- [x] Kinematic path untouched; determinism step byte-identical
- [x] `--dry-run` never arms anything
- [x] Tests: flag on arms the two aircraft and not the towers; flag off arms nothing

## `docs/arena.md` §3 was rewritten, and this is why

It said the run *"does not arm or launch anything"*, *"posts nothing to the tracks API"* and *"runs no vision"*, and concluded **"a judged run made with this command scores nothing"**. Two of those stopped being true with #134 and the third stops here.

That section is what someone reads *before the judged run*. Leaving it telling the operator the opposite of the truth, tonight, was the more dangerous option than editing a doc in a feature PR. It's now a table of preconditions plus an honest list of what still isn't automatic — tower aiming beyond waypoint-as-look-at, and tower siting.

## What is still not verified

**None of this has run against the real arena.** It cannot be, from the machine it was written on — there's no arena reachable here. The tests drive `cmd_run` against an arena-shaped fake that records what it was asked to launch; they prove the wiring and the gating, not that ArduPilot arms.

The first live run is the test. `docs/arena.md` §3 is the command.

---

## Review round two

Every finding is addressed. The merge hazard resolved itself when `main` was merged in twice: the merge base is now `main`'s tip, `AXES` still comes from `whiteout.score`, and #133's scorer is intact.

**`--launch-alt` is bounded.** `argparse` took `-50`, `0`, `nan` and `inf` as floats and forwarded every one into `MAV_CMD_NAV_TAKEOFF`. It must now be finite and positive — the bound `SearchParams` already puts on `search_alt_m`, which is this flag's default. No ceiling: that would be a number invented in the CLI rather than one the arena or the policy states. The check runs whenever `--arm` is passed, **including on the kinematic transport**, because a rehearsal is where a typo in this flag should be found rather than the one live run.

**"launched" claimed more than the code knows.** `arm_and_launch` pipelines the arm and the takeoff without reading `COMMAND_ACK`, on purpose — the arming window is about three seconds — so a pre-arm rejection and a climb are indistinguishable from here. The line now reads `run: arm and takeoff to 120 m sent to quadcopter, fixed-wing` and points at the fleet rows and gzweb for what actually happened. `docs/arena.md` no longer tells the operator that line is how they know.

**An `OSError` no longer ends the run.** Only `TransportError` was caught per asset, so a pymavlink socket that dropped mid-roster killed the episode before tick 1 — the opposite of what "one asset that will not arm does not end the episode" promises.

**Tilt is never commanded, and the docs say so.** `_aim` sends servo 1 and nothing else, and `set_servo` has no caller repo-wide, so a tower holds whatever tilt it booted with — and a level tower is blind inside `height / tan(vfov/2)`, 695 m for the taller mast.

The unreachable `not callable(launch)` guard is gone; `cmd_run` casts on the arena branch, where the factory has just built an `ArenaTransport`.

**Nine tests, each verified to fail against the previous commit.** One pins something worth knowing: `argparse`'s negative-number matcher is `^-\d+$|^-\d*\.\d+$`, so `--launch-alt -50` is handed over as a value and refused here, while `--launch-alt -inf` is refused by the parser first as a missing argument. Both are refused, by different code.

Gate: `1015 passed, 12 skipped`; `lint`, `format`, `build`, `smoke` and `determinism` all pass, determinism byte-identical. `typecheck` fails locally on `whiteout/vision/imagery.py:267` — an unused `type: ignore[import-not-found]` for `cv2`, which reproduces with this branch's changes stashed and is an artefact of this machine having OpenCV installed. Left alone: the ignore is needed wherever `cv2` is absent, and the file is not in this diff.


Ollienel777 (commented):
Verdict: report findings — don't merge yet

Gate is green (963 passed, lint/format/build/smoke/determinism all pass, determinism byte-identical). The design is careful and the opt-in gating is right: --arm is off by default, --dry-run wins over it and says so, towers are filtered by class rather than having their refusal caught, and the kinematic path is untouched. But there's a merge hazard and three correctness gaps on the one code path that moves real vehicles.

Findings
whiteout/cli.py:29 — critical. Merging main conflicts in all three files (12 markers, 6 in cli.py). The branch carries #134 as real commits, main carries it squash-merged, so the merge base is 8552e50 ([#128](https://github.com/Ollienel777/HTN_WHITEOUT/issues/128)), before [#134 · feat(infra): open the cameras…](https://github.com/Ollienel777/HTN_WHITEOUT/pull/134). The cli.py conflict pits the head's hardcoded AXES tuple against main's from whiteout.score import AXES, ScoreError, score_episode, ... — resolving in the head's favour silently reverts [#133 · feat(score): derive the axes…](https://github.com/Ollienel777/HTN_WHITEOUT/pull/133)'s scorer. gh reports MERGEABLE, which disagrees with what git actually does and was reported while the re-triggered gate was still running.
whiteout/cli.py:511 — medium. --launch-alt is unvalidated: verified that -50, 0, nan and inf all parse and are forwarded unchecked into MAV_CMD_NAV_TAKEOFF (arena.py:551). The repo hand-diagnoses a negative --seed and refuses p_max=1.0; this is the one flag that commands vehicle motion.
whiteout/cli.py:258 — medium. "launched" means commands sent. arm_and_launch is fire-and-forget with no COMMAND_ACK read; its own docstring says the result is read from the pose stream and nothing does. An ArduPilot pre-arm rejection prints run: launched quadcopter, fixed-wing to 120 m with both on the ground — and docs/arena.md:155 tells the operator that line is how they know.
docs/arena.md:167 — medium. Says command turns a waypoint into "servo 1 and servo 2". _aim sets servo 1 only; set_servo has exactly one call site repo-wide. Tilt is never commanded.
whiteout/cli.py:258 — low. Only TransportError is caught per asset; an OSError from the pymavlink socket escapes and kills the run before tick 1, contradicting "one asset that will not arm does not end the episode".
whiteout/cli.py:245 — nit. The not callable(launch) branch is unreachable (arena-only call site, ArenaTransport always defines it) — same shape as the dead except VisionError removed from #134.
docs/arena.md:141 — nit. The arming row omits WHITEOUT_TRANSPORT=arena, which the row above states for waypoints.

Ollienel777 (approved):
lgtm!!!

Issue #90, opened 2026-09-19
vision: the detector has never been measured on compressed imagery, and it does not survive it


Ollienel777:
Found while fixing #84's review findings. Not fixed there, and it is bigger than the finding that led to it.

## What

`whiteout/vision/scene.py` adds Gaussian sensor noise as its **last** render step, after fog and after blur. So adjacent pixels differ by exactly the white noise `_row_background`'s scale estimator assumes. That is the one property real imagery is guaranteed not to have, and the arena publishes JPEG.

`scripts/jpeg_quantisation.py` (added in #84) puts the generator's frames through baseline luma quantisation — forward 8x8 DCT, standard luminance table, inverse. Reproduce with:

```
python scripts/score_detector.py --synthetic --per-camera 24 --seed 11
python scripts/score_detector.py --synthetic --per-camera 24 --seed 11 --jpeg 75
```

## What it costs

Detection rate / false positives per empty frame, 24 frames per camera, seed 11, on the code as it stands after #84:

| scene | uncompressed | quality 75 |
|---|---|---|
| `CLEAR` | 0.58 / **0.02** | 0.31 / **0.52** |
| `FOGGY` | 0.00 / **0.00** | 0.23 / **0.58** |
| `HEAVY_FOG` | 0.00 / **0.00** | 0.19 / **0.56** |

Heavy fog goes from correctly silent to reporting a vessel on more than half of the frames that have none. Every one of those posts a lat/lon. The false-positive rate is 25x the uncompressed one and recall is *lower*, so this is not a threshold sitting in the wrong place — the statistic itself stops meaning anything.

The mechanism is the one #84 fixed a floor against: the lag-1 scale estimate falls to a third of the noise actually present at quality 75 and onto `min_sigma` at quality 60, so every gate quoted in sigmas opens at once. `min_depth_counts` is one floor that cannot be opened that way, and on the uncompressed corpus it costs nothing, but it is not enough on its own — the table above already has it.

## Two candidate fixes, both measured, neither shippable on this evidence

[...]

PR #142, opened 2026-09-20
fix(vision): raise the one gate a collapsing sigma cannot open

chessdudeo7:
Refs #90 — the measurement half, and one parameter. **The issue stays open**: this does not settle it, and the reason is at the bottom.

## What was wrong

The arena publishes JPEG. At `min_depth_counts=4` the detector reported a vessel on **more than half the empty frames** of a quantised corpus — 25× the uncompressed rate, *with lower recall*. That combination is the tell that this was never a threshold sitting in the wrong place: the statistic itself had stopped meaning anything.

## Two candidate repairs ruled out first

#90 proposed estimating the noise at several pixel separations and keeping the largest. **Measured, and it does not work** — the scale estimator is not the dominant cause:

| scale estimate | uncompressed recall / fp | q75 recall / fp |
|---|---|---|
| lag 1, per row (shipped) | 0.583 / 0.021 | 0.312 / 0.521 |
| lag 2, per row | 0.562 / 0.000 | 0.354 / 0.521 |
| lag 8, per row | 0.396 / 0.000 | 0.479 / 0.604 |
| lag 16, per row | 0.062 / 0.000 | 0.396 / 0.104 |
| lag 1, **global** | 0.521 / 0.000 | 0.312 / 0.500 |
| lag 8, **global** | 0.521 / 0.000 | 0.417 / 0.646 |

I expected lag 8 to win — JPEG quantises within 8×8 blocks, so pixels 8 apart sit at the same phase in *adjacent* blocks and should stay decorrelated. It doesn't. Nor does making the scale global, which was the principled answer to #90's "raising it per-row moves the argmax". **No lag and no scope recovers the false-positive rate.** That rules out the whole family and saves the next person the hour.

## What does move it

The one gate quoted in **counts** rather than in sigmas, because a collapsing denominator cannot open it. Three seeds, 24 frames per camera:

| seed | counts | uncompressed recall / fp | q75 recall / fp |
|---|---|---|---|
| 11 | 4 | 0.583 / 0.021 | 0.312 / **0.521** |
| 11 | **12** | 0.500 / 0.000 | 0.438 / **0.104** |
| 23 | 4 | 0.646 / 0.000 | 0.312 / **0.521** |
| 23 | **12** | 0.542 / 0.000 | 0.542 / **0.250** |
| 47 | 4 | 0.604 / 0.042 | 0.354 / **0.542** |
| 47 | **12** | 0.479 / 0.000 | 0.375 / **0.062** |

On the compressed corpus 12 **dominates** 4 on every seed — more detections *and* two to eight times fewer false alarms. It costs about a tenth of the recall on uncompressed frames, which the run will never see, because the arena does not serve uncompressed.

A false positive posts a lat/lon, and *accuracy* is scored. That is the direction to err in.

## Why the issue stays open

**This is not tuned against Dominion Dynamics' encoder**, only against `scripts/jpeg_quantisation.py` — no chroma, no subsampling, no entropy coding.

**And the gate is not monotone.** Raising this floor deletes rivals before `runner_up` is read, so a tighter gate can *raise* the detection rate: 4 → 10 does exactly that on seed 11 (0.312 → 0.667). **Part of the gain above is that defect rather than a stricter detector**, and I would rather say so than present 12 as a clean win. The structural repair — snapshot the rival field before the gates — still halves recall on clean frames, and choosing between that and a non-monotone gate still needs #63.

Both caveats are in the `min_depth_counts` docstring, not just here.

## The test that failed, and why I changed it rather than the code

`test_the_counts_floor_does_not_cost_a_hull_that_is_really_there` bracketed the floor with four constants — 3, 4, 6, 8 — and went red because 6 and 8 are now below it.

That guard is exactly right in intent: *"the floor has to sit under a real hull, or it is just a blindfold."* But it failed for the one reason a guard must never fail for — **it was bracketing a value the code no longer held, so it reported a blindfold where there was a working gate.** The ladder is now written relative to `DEFAULT_PARAMS.min_depth_counts` (0.5×, 0.75×, 1.25×, 1.5×), so it keeps asking its question at whatever the floor becomes next.

## Also

`scripts/score_detector.py` now says, **without** `--jpeg`, that its rates are numbers about uncompressed frames. #90 asked for this and it is the one part of the issue that needed no judgement: the headline false-positive rate read as a statement about the detector when it was a statement about frames the run will never see.

## Verified

Gate green, 8/8, determinism byte-identical.


Ollienel777 (commented):
Yes, mergeable — but one gap worth closing first

MERGEABLE / CLEAN, gate SUCCESS. It's 6 commits behind main but merges with no conflicts; post-merge locally: 1050 passed, 12 skipped, lint/format/build/smoke/determinism all pass.

The evidence is real, and I checked it rather than taking it. I reproduced four rows of the docstring's tables exactly — seed 11 at counts 12 under quality 75 gives detection_rate 0.4375 / false_positive_rate 0.10417 against the claimed 0.438 / 0.104, and seed 47 matches at both 4 and 12. The headline holds: at the old floor of 4, JPEG quantisation pushed the false-positive rate on empty frames to 0.521–0.542, and 12 genuinely dominates it on compressed frames — more detections and five to eight times fewer false alarms. The docstring's candour is earned: it says plainly that 12 isn't the optimum, that part of the gain is a runner_up interaction rather than a stricter detector, and that #90 stays open.

No fixture risk. I checked the thing that bit #125: whiteout/shadow.py ([#149 · feat: a shadow vessel for the…](https://github.com/Ollienel777/HTN_WHITEOUT/pull/149)) drives sightings off the belief model, not detect_vessel, so this parameter can't reach the demo episode. Confirmed empirically — the PR's run --seed 7 --ticks 400 is byte-identical to main's, and the committed fixture's difference from both is the pre-existing cross-machine numpy drift that fixtures/episodes/README.md documents.

The score_detector.py banner logic is also sound: if --jpeg → elif --synthetic → else is exclusive and exhaustive, and --synthetic --jpeg correctly takes the first branch.

Findings
tests/test_vision_detect.py:364 — medium. I swept the shipped value and ran the whole suite each time: min_depth_counts = 4.0 → 1050 passed, 12 skipped. Setting the parameter back to exactly the value this PR exists to move away from is completely invisible. 8.0 passes too; only 20.0 (2 failures) and 30.0 (7) bite. The relative ladder is the right call for the reason the docstring gives — a fixed ladder was silently bracketing a stale value — but it pins discrimination, not position, and score_detector.py isn't in the gate, so nothing guards a parameter the PR itself calls "a cliff edge". A corpus-at-quality-75 test asserting the false-positive rate stays under ~0.2 would fail at 4.0, costs seconds, and can sit beside the ladder.
whiteout/vision/detect.py:185 — design, not a defect. The choice of 12 over 11 is argued solely on accuracy being scored. But since [#133 · feat(score): derive the axes…](https://github.com/Ollienel777/HTN_WHITEOUT/pull/133)/[#146](https://github.com/Ollienel777/HTN_WHITEOUT/issues/146) the scorer has five axes, and recall drives detection_speed and tracking_duration directly with no downstream recovery — while a false positive must still clear MotionGate ([#110](https://github.com/Ollienel777/HTN_WHITEOUT/issues/110)) and TrackHold._could_have_got_there ([#114](https://github.com/Ollienel777/HTN_WHITEOUT/issues/114)) before it can be POSTed. On seed 47, counts=11 is 0.625 recall / 0.125 fp against 12's 0.375 / 0.062. The asymmetry argues the fp side is partly absorbed and the recall side isn't.

Issue #143, opened 2026-09-20
feat(transport): a tower never tilts, so it is blind inside its own foot


Ollienel777:
Split out of #68 and #137, both of which assume this already works. It does not.

## What

`_aim` should pick a **tilt** as well as a pan, and send it on `servo set 2`.

## Why

`whiteout/transport/arena.py`'s `_aim` computes a bearing and sends
`servo set 1` only. `set_servo` has exactly one call site repo-wide — that
one — so **servo 2 is never commanded anywhere in the product**. A tower holds
whatever pitch it booted with, for the whole episode.

That is not free:

1. **It costs measured coverage.** A level tower is blind inside
   `height / tan(vfov/2)` — **358 m** for tower-1 at 116.8 m and **695 m** for
   tower-2 at 226.5 m, against the 36.1° vertical field in
   `whiteout/vision/camera.py`. `scripts/tower_siting.py` (#135) scores the
   taller mast at **18.7% level and 26.1% sweeping tilt**: the blind disc is
   worth **7.4 points of channel coverage** on the one asset that is free,
   permanent and never tires.
2. **The siting numbers we are presenting assume it.** `SWEEP_TILTS_DEG`
   sweeps seven tilts, and its comment says "tilt is commanded — `servo set
   2`". It is not. Today the honest column is the `--level` one, which is the
   version that ticket deliberately argues against. Either drive tilt or
   restate the result.
3. **Aiming at a lat/lon without a pitch is half an instruction.** The policy
   hands `_aim` a point on the water. Bearing plus range plus height
   determines the depression angle; discarding it means the camera looks at
   the bearing of the target and the elevation of the horizon.

## The pose side, which has to move with it

`_pose` reads pitch from `ATTITUDE` and reports it honestly (`None` when
there is none, per #99), and the non-detection update erodes water using that
pitch. So the belief field is not currently lying — it is eroding the level
band the tower really holds. **That stays true only if the tracker's
`ATTITUDE` follows its tilt servo.** `MAV_CMD_DO_SET_SERVO` writes a raw
pulse and bypasses the tracker's own pitch control, so it is an open question
whether `ATTITUDE.pitch` moves with it at all. Verify against the arena
before trusting either path.

Related: `whiteout.vision.tower.tower_camera_pose` turns the PWM pair into a
`CameraPose` and is exercised only by tests — nothing in the product calls
it. If `ATTITUDE` turns out not to track the servo, that function is the
[...]

Issue #137, opened 2026-09-20
feat: a tower turns to look where it is told


Ollienel777:
Split out of #68. The aiming half, and specifically the part the **fake** gets wrong.

## What

`KinematicTransport` ignores tower intents. Make a tower turn.

## Why

The policy is already aiming them. Over the committed 400-tick episode it issues **16 distinct aim points to tower-1 and 14 to tower-2** — and the transport discards every one. Tower headings are constant for the whole episode at the values `connect` hardcodes:

```
distinct tower headings over 400 ticks: {tower-1: [106.49], tower-2: [282.26]}
```

Three consequences:

1. **It is a fake-vs-real divergence.** `whiteout/transport/arena.py` already does this properly: `command()` routes a tower intent to `_aim()`, which turns a bearing into `servo set 1` / `servo set 2`. The arena pans; the fake does not. The demo under-represents the system we built.
2. **Two of four assets stare at one patch for 200 seconds.** With the non-detection update live (#13), their footprints saturate in the first few ticks and then contribute nothing.
3. **Measured, it costs peak probability.** Feeding the commanded bearings back into the fake over the same episode: `peak_p` 0.00429 → 0.00954, a 2.2x improvement in how sharply the field concentrates. Entropy 5.9286 → 5.8485.

It also matters for #20: a canvas drawing frozen tower footprints looks like a bug, where sweeping ones *are* the collaboration beat — two immobile assets working the ends while the aircraft take the middle.

## What is already done

Do not rebuild these:

- `arena.py` has `aim()`, `set_servo()` and `scan()`
- `arena.py`'s `command()` already branches on `cls == "tower"` and calls `_aim`
- Tower siting is data on the arena side, loaded from a file

## What is missing

- `KinematicTransport._advance` leaves a tower's `heading` alone. It should turn toward the bearing from its own position to the commanded `target_lat/lon`, and **report the heading it turned to** — the non-detection update reads `pose.heading` as `yaw_deg`, so a tower that turns without saying so erodes the wrong water.
- A turn rate. An AntennaTracker does not snap; pick one, name it on `KinematicAsset` beside `cruise_mps` and `climb_mps`, and say what it is a guess about.

## Acceptance

- [ ] A commanded tower intent changes that tower's reported `heading`
- [ ] The heading is **degrees**, like every other transport (this bit me once already — see #123)
- [ ] A tower still never changes `lat`/`lon`: it is aimed, never sent
[...]

PR #145, opened 2026-09-20
docs(transport): _aim sends a pan, and nothing tilts

### LINKED ISSUE #144: docs(transport): _aim's docstring claims a tilt command the code never sends
`whiteout/transport/arena.py`'s `_aim` says:

```
"""Point a tower's camera at a lat/lon, via servo 1 and servo 2."""
```

It sends servo 1 and nothing else. `command()`, one method above, repeats the
claim — "turned into pan and tilt". `set_servo` has exactly one call site
repo-wide, and it is that pan, so **servo 2 is never commanded anywhere in the
product**.

`docs/arena.md` carried the same claim and was corrected in #140, so the doc
and the docstring now disagree and the docstring is the wrong one. #140 left
`arena.py` alone deliberately, to keep its diff to the three files it owns.

Whether tilt *should* be driven is #143. This ticket is only the docstring: a
comment that describes a command the code does not send is how the same
mistake gets made again — it is already in #137's body and in
`scripts/tower_siting.py`'s `SWEEP_TILTS_DEG` note, both of which took it from
here.

## Acceptance

- [ ] `_aim`'s docstring says it sends a pan on servo 1 and that tilt is not
      commanded, with the consequence — a level tower is blind inside
      `height / tan(vfov/2)` — rather than only the fact
- [ ] `command()`'s docstring no longer claims tilt
- [ ] It points at #143 so a reader who thinks this is a bug finds the ticket

Ollienel777:
Closes #144

`_aim` said "via servo 1 and servo 2". It sends servo 1. `set_servo` has one
call site repo-wide — that pan — so **servo 2 is never commanded anywhere in
the product**, and a tower holds whatever pitch it booted with. `command()`,
one method above, repeated the claim.

`docs/arena.md` was corrected in #140, which deliberately did not touch
`arena.py`, so the doc and the docstring disagreed and the docstring was the
wrong one.

The new text says what it costs, not just what it does — a level tower is
blind inside `height / tan(vfov/2)`, 358 m from tower-1 and 695 m from
tower-2, which `scripts/tower_siting.py` prices at 7.4 points of channel
coverage on the taller mast — and points at #143, so a reader who thinks this
is a bug finds the ticket rather than filing it again.

**Whether tilt should be driven is #143**, opened alongside this. The short
answer is yes, and the part that needs the arena is the question of whether
`ATTITUDE.pitch` follows `servo set 2` at all: `MAV_CMD_DO_SET_SERVO` writes a
raw pulse and bypasses the tracker's own pitch control, and the belief field's
non-detection update erodes water using the reported pitch. Not widened into
here.

Two other copies of the same claim are noted on #143 rather than fixed here,
because both become true if #143 lands: #137's body, and
`scripts/tower_siting.py`'s `SWEEP_TILTS_DEG` comment — whose seven-tilt sweep
is what produces the 26.1% figure the siting argument rests on.

## Gate

| step | |
|---|---|
| `ruff check whiteout tests scripts` | pass |
| `ruff format --check whiteout tests scripts` | pass — 75 files |
| `mypy whiteout` | one error, pre-existing and environmental: an unused `type: ignore` at `whiteout/vision/imagery.py:267`, which appears only when OpenCV is installed locally. Not touched by this PR |
| `pytest -q -m "not slow"` | pass — 998 passed, 12 skipped, 1 deselected |

Comments only; no behaviour changes.


Ollienel777 (commented):
lgtm

PR #149, opened 2026-09-20
feat: a shadow vessel for the fake fleet to find

Ollienel777:
The kinematic demo was a search that never succeeded. `score` said so plainly:

```
detection_speed:    not detected — no record of 400 carries a contact
tracking_duration:  not detected — there was never anything to hold
```

Every part of the chain — detector, projection, track hold, poster — was merged and correct. Nothing was there to find. **"Found and held" is half the project's one-liner and the demo showed the other half.**

## Now

```
detection_speed:   0.8947    first contact at t=21.0 s
tracking_duration: 0.6925    277 of 400 ticks held, 138.2 s of 199.5 s
search_efficiency: 0.9833
coverage:          1.0000
total              0.8926    over 4 of 5 axes, against 2 before
```

The viewer shows it: `CONTACTS 1 · Sierra One · tracked · 1.00`.

## Why it is not in `whiteout/transport/`

The seam, and the seam is right. That package is adapters only, and the no-leak guard fails any module there naming a ground-truth concept. **A vessel whose real position is known is exactly that**, and a `kinematic` transport carrying one would be a fake that knows something the real link never can.

So it reaches the loop the way the arena's cameras do — as a `SightingSource` — and `cli.py` picks one or the other. Nothing downstream can tell which it got, which is the property that makes a rehearsal worth anything.

## The sensor model is the belief field's own, deliberately

An asset sees the vessel when `non_detection_likelihood` — the same function the field erodes with — says it would.

**If the model producing sightings disagreed with the one spending them, the fake would be tuning the policy against a world that does not exist.** Belief would be erased from water a sighting could still come out of, or held over water no camera could have covered, and every number the demo quotes would be a number about the disagreement.

The cost is stated in the module docstring rather than left to be found:

- **This fake cannot surprise us.** It will never produce the detection the belief model says is impossible, so it tests the loop's *plumbing* and not the sensor's fidelity. The arena tests that; `scripts/score_detector.py` measures it.
- **No false alarms.** `detect.py` has a whole apparatus for refusing a floe shadow that looks like a hull, and none of it is exercised here. A clean run says nothing about the false-positive rate — #90 is where that lives.

Speed is **3.0 m/s, read off the arena's own `.env`** (`SHIP_SPEED=3.0`, `SHIP_MOVING=1`, `SHIP_START=random`) — not the 2.96 m/s we once measured, which was a chord across an arc and so a lower bound by construction.

## Two tests changed, and why

`test_score.py` had two tests asserting the demo fixture carried no contacts. The property they were really testing — *an absence is not a measured zero* (R1-M1) — is about the **scorer**, so it moved to a synthetic contact-free log. The fixture gained the inverse assertion: a `demo.jsonl` that goes back to "not detected" means the fleet stopped finding things, **which is the state it spent a day and a half in without anything failing**.

## Gate

`1045 passed, 12 skipped`, lint and format clean, determinism byte-identical at seed 7, fixture regenerated.

`--no-vessel` runs the old pure search, so the before/after is reproducible rather than asserted.

## Known limitations

Beyond the two above: the fix error (`DEFAULT_FIX_SIGMA_M = 60 m`) is a standing assumption until the projection's real error is measured against the arena, and the vessel's turn rate is chosen rather than observed.


Ollienel777:
Review finding fixed in `e8cf375`.

**The East-West fix error was 5.7 m against the 60 m the field documents** — the offset was divided by the ratio between the two axes' metres-per-degree where it needed multiplying, so it missed by that ratio squared (10.4x at 71.99 N). Measured back through `whiteout.geo`: 59.7 m North-South, 5.7 m East-West.

Bellot Strait runs East-West — the two tower stations differ by 0.12 deg of longitude against 0.01 deg of latitude — so that was the **along-channel** axis, which is the one `TrackHold.course()` measures a bearing over. `DEFAULT_FIX_SIGMA_M`'s docstring says a sighting that landed on the target every time "would make the track hold look better than it is", and along the channel it nearly did.

There are now two constants, one per axis, and each offset is divided by the scale of the axis it is on. `_offset` takes that scale as an argument rather than closing over one, so a call site cannot hand it the wrong axis. `_LON_PER_LAT` is gone; a single ratio between two axes is the shape that got this wrong once.

**The first test I wrote for it passed against the defect**, which is worth recording because it is the same trap as #139's R1-M3. It called `_offset` directly with a scale it computed itself, so it pinned the helper while the call sites stayed wrong. The version that landed measures the spread of the positions `sightings` actually reports, in metres rather than degrees, with the clock held still so the vessel does not move between draws. Two mutations fail it: the original expression verbatim, and the naive slip of passing the latitude scale for longitude.

**The fixture is regenerated and the demo is worse, correctly:**

| | before | after |
|---|---|---|
| `detection_speed` | 0.8947 | 0.8947 |
| `tracking_duration` | 0.6925 | **0.5800** |
| `search_efficiency` | 0.9833 | 0.9837 |
| total | 0.8926 | **0.8646** |

232 of 400 ticks held rather than 277. The hold was being helped by an along-channel error an order of magnitude smaller than the one this module says it injects. The numbers in the PR description above are the pre-fix ones.

Gate: lint and format clean, `1046 passed, 12 skipped, 1 deselected`, wheel built, smoke ran, determinism byte-identical at 1310885 bytes.

PR #150, opened 2026-09-20
fix(tracks): a time floor on the course baseline, so noise is not speed

Ollienel777:
Found by the shadow vessel (#149), and it could not have been found before: until this morning no kinematic run produced a second fix.

## What

`TrackHold.course()` required a **15 m baseline and no minimum time**. That threshold was sized against a fix error of about 3 m — the comment says so — and nothing has measured the real one. The projection's flat-water model is 10% wrong at its own range bound, 200 m at 2 km, so an error well above 15 m is the *expected* case at range, not the pessimistic one.

When it is, every adjacent pair clears 15 m on noise alone. The walk back through the history therefore stops at the **newest** qualifying pair — the noisiest one available — and divides that noise by a one-second gap.

## Measured against a known 3.0 m/s

```
                      before          after
worst reported speed  81.2 m/s   ->   16.5 m/s
median                 ~9  m/s   ->    4.7 m/s
```

**81.2 m/s is twenty-seven times the truth**, posted on a field `ARENA.md` §5 carries in its own example payload.

Requiring **10 s as well as 15 m** fixes the catastrophic case. Ten seconds is thirty metres of travel at 3 m/s, so the signal beats a two-sigma fix error rather than merely exceeding a floor.

## What is left, and why it is not a threshold to tune

4.7 m/s median against 3.0 m/s truth is still high, and the residual is **structural**: a chord across ten seconds of a turning vessel under-measures the path while fix error inflates it, and those do not cancel. A longer baseline trades one against the other rather than removing either. What is gone is the case where the number was nonsense.

## The test

Held against a vessel that **does not move at all** — the sharpest version, since every metre of reported speed is error.

The ellipsoid guard refused my first attempt, which wrote `60.0 / 111_000.0` to turn metres into degrees. It was right: that is the second converter the guard exists to stop, and a test is no more exempt than a module. The jitter is drawn in metres and converted by `local_to_geodetic`.

## Two process notes

**A red suite reached a commit.** The command was `ruff && ruff && pytest -q | tail -3 && git add && git commit`, and the pipe makes the exit status `tail`'s, which always succeeds. Running the suite to a file and reading the tail separately is the shape that does not lie; the second commit records it.

**This PR was eight minutes late** because I ran `until gh pr checks | grep -qE "pass|fail"` *before* creating the PR and without a number. With no PR on the branch that command errors every time, so the loop could never exit. A wait-for-CI loop belongs after `gh pr create`, and with the number.

## Gate

`1046 passed, 12 skipped`, lint and format clean.


Ollienel777 (commented):
Not yet — don't merge this one without resolving the conflict deliberately. The change itself is right, and worth more than the PR claims, but the merge is booby-trapped.
Findings

whiteout/shadow.py:250 — branch predates #149 and carries the pre-fix shadow.py; an add/add conflict whose natural resolution silently reverts the longitude fix.
whiteout/tracks/maintain.py:116 — the measured figures (81.2 m/s etc.) were taken against the buggy sensor and don't describe main; they land as permanent comments.
whiteout/tracks/maintain.py:251 — course()'s "returns (None, None) when…" list omits the new first-10s window.
The merge hazard
I attempted the merge to confirm the shape:

CONFLICT (add/add): whiteout/shadow.py
CONFLICT (add/add): tests/test_shadow.py
CONFLICT (content): fixtures/episodes/demo.jsonl
An add/add conflict presents the whole file on both sides. Line 250 of this branch's copy still reads self._offset() / _LON_PER_LAT — the 10× bug. And tests/test_shadow.py is conflicted too, so the same "keep mine" resolution drops the test that would catch it. Nothing would fail.
Resolve by taking main's side wholesale for whiteout/shadow.py and tests/test_shadow.py, keeping only whiteout/tracks/maintain.py and tests/test_tracks_maintain.py from this branch, and regenerating the fixture with the documented command rather than merging it by hand.
The change is better than advertised
I re-measured on current main, same seed-7 episode, capturing every non-null speed_mps handed to the poster:

main as-is (no time floor)   median 14.62   p90 149.75   worst 188.95
with this PR's 10 s floor    median  6.76   p90  13.52   worst  15.61
So the defect is 189 m/s, 63× truth — not the 81.2 the PR reports — and the fix takes the worst case to 15.6. The PR's numbers came from the branch's own sensor, whose along-channel error was a tenth of what it should be. Both sides of the before/after are understated.
That matters beyond bookkeeping, because the numbers are the argument for 10 s: "thirty metres of travel at 3 m/s, so the signal beats a two-sigma fix error." At the corrected 60 m isotropic sigma, two-sigma on the difference between two fixes is ~170 m, which 30 m of travel doesn't beat — and the measured 6.76 m/s residual against 3.0 m/s truth is consistent with that. 10 s may still be the right threshold, but the stated justification no longer holds, and the figures are being committed as module comments.
I did verify the test earns its place: removing the time floor fails test_a_noisy_fix_is_not_read_as_speed. And the two process notes in the PR body — the | tail -3 swallowing pytest's exit status, and the wait-for-CI loop running before gh pr create — are both real and well spotted.

Ollienel777:
All three fixed. `MERGEABLE`, gate green, `1047 passed`.

## 1 — the merge hazard

Resolved exactly as prescribed, and it was as booby-trapped as described. Both `whiteout/shadow.py` and `tests/test_shadow.py` conflicted **add/add**, which presents the whole file on each side, and this branch's copy still read `self._offset() / _LON_PER_LAT`. A "keep mine" resolution would have reverted #149's follow-up fix *and* dropped the test that catches it, silently.

Took `main`'s side wholesale for both, kept only `tracks/maintain.py` and `tests/test_tracks_maintain.py` from here, and regenerated the fixture with the documented command. Verified the committed fixture is byte-identical to what that command produces rather than assuming it.

## 2 — the figures, which were the argument

You are right that they were measured against the branch's own broken sensor and were going in as permanent comments. Re-measured on `main`, seed 7, 60 m isotropic, against a known 3.0 m/s — and I swept the threshold rather than just re-measuring the two points:

```
time floor   median     p90    worst   posts
none          14.62  145.80   188.95      16
10 s           6.76   11.71    15.61      16
20 s           4.50    6.58     9.58      15
30 s           4.05    5.73     7.56      13
```

Your numbers reproduce exactly. **188.95 m/s, 63x truth** — not 81.2 and 27x.

## 3 — the justification, which mattered more

Also right, and this is the part I would have shipped as a plausible-sounding falsehood. *"Thirty metres of travel at 3 m/s, so the signal beats a two-sigma fix error"* — two independent 60 m fixes differ with a sigma near 85 m, so two-sigma is ~170 m and 30 m is nowhere near it.

The real relationship is that **the noise term falls as 1/T while the signal does not**: at 10 s that is ~8.5 m/s of noise on 3.0 of truth, which is precisely the 6.76 median. Beating it outright needs T >= 57 s, most of an episode. The comment says that now, and names the claim it replaces.

## Why still 10 s and not 30

The table plainly favours 30 s, so the reason is in the comment rather than left implicit: **that table is measured against `shadow.py`'s vessel, whose turn rate is chosen rather than observed.** A chord under-measures a turning path, so the longer the baseline, the more of the answer is a statement about how sharply the target turns — the one property of the real vessel nothing here knows. Ten seconds keeps the catastrophic case out while leaning least on it.

That is the same trap `shadow.py`'s own docstring warns about — tuning against a world that does not exist — and 30 s would have walked into it for a better-looking number.

## 4 — the docstring omission

`course()`'s "(None, None) when..." list is three cases now: the first fix, a stationary vessel, and **the first ten seconds of any track**, which is new with the floor and is its price. A fix with no course is worth more than a course that is fix noise.

---

Worth noting what found this: the 189 m/s defect has been on `main` since the track hold was written, and nothing could see it until there was a second fix to difference. The shadow vessel produced one this morning.

